Issue #45, opened 2012-10-17
Digest Approval Notification Emails


paulca:
I think a daily email, or twice daily email showing that there are submissions awaiting approval would be great.

matschaffer:
Re-opened for feature review

sasharevzin:
Is it something that still needs to be implemented?

auspicacious:
@sasharevzin yes, this would be appreciated.

PR #47, opened 2012-10-30
Digest admin email

rknLA:
Addresses #45 

The idea is that the DigestEmailer is a sort of singleton.

Each import that's uploaded gets added to the current digest.  If no current digest exists, one is created with a send date 24 hours in the future, and then the import is added.  Imports uploaded within that window get added as expected.

Worst case scenario is that it takes 24 hours for a single import approval email to propagate.  Best case, there are a handful of imports in one email.

Testing is a bit minimal, since it's not really clear to me how to test the emailing mechanism when it's scheduled for the future (`Delayed::Worker.new.work_off` doesn't do anything if the job is scheduled for 1 day from now unless I let the test run for a day).

We may also run into a bug where calling `.deliver` on a notification mailer inside a delayed job could cause it to send twice.  The example case is when `.delay` is used in the method call chain, so it's not clear if we'll see that symptom until we actually start sending email through this process.

@paulca please review and comment.


matschaffer:
Closing this out given age & need of update, etc. I'm also thinking when we pick this back up we may want to instead have a 12 hour cron that sends the digest for anything pending approval. It think it could be a little simpler to implement and would meet @nokton's "approve with the sun" feature request.

PR #162, opened 2014-09-16
Fixdrive button so admins can force imports which have some md5 collisions

robouden:
if one of the measurements has the same MD5 the whole drive will not be going live..

I went around that by ignoring the MD5 when finalising.. maybe not correct way but it worked..


seanbonner:
This is old, but is it valid? I'll merge it if so.

nokton:
Rob should check what was meant with “fix drive” functionality. 

We have a bunch of drives that are “stuck”, and need some script so that the drives can be worked on. Not sure, but if this is for that, then we should keep this item.

Pieter

robouden:
"Fixed drive" is an option for the admin to insert measurements in the
database by by passing the MD5 checksum. Some  imports are "hanging"
because the same md5 checksum exists. The test was made and is working on
dev.safecast.org. Test was check by other persons and agree upon.


rob



Regards,
Rob Oudendijk Yuka Hayashi
http://yr-design.biz   http://oudendijk.biz  http://about.me/robouden
tel +81 80-22605966  Skype: robouden  Facebook:robouden
<http://on.fb.me/QeKw2P> twitter:robouden <http://bit.ly/RAiSTC>

nokton:
Can we move it to production so we can close this item?

Pieter

matschaffer:
Seems like there's some stuff in here not related to fix drive (devices sql dump, docs graphiviz output, pgpass file). Might be good to clean that up before a master merge.

nokton:
Hi MArk,

Sorry that was a bit too cryptic for me ;) - what is the downside/dependencies if we don’t clean that up and go for a release (at least to the Dev environment) 

Pieter

matschaffer:
If it's indeed "cruft" I'd say the main downside is cognitive load.

For example, if a developer needs to track down a bug related to these changes, they'll need to take time qualify what's part of the feature and what isn't.

Also makes the repo a little larger, and pgpass could be a security concern.

I don't think any of this is a massive problem, of course, but one of those "good housekeeping" sort of things that we should avoid if possible.

nokton:
Hi Mat,

Got that - my preference is to parse things into the dev server as early as possible so we can try it out. Also as we have not done this for a while, will be great to see we have the devops in place to get things from from coding to dev to prod.

Pieter

app/models/bgeigie_import.rb
```diff
@@ -88,6 +88,37 @@ def finalize!
...
+     insert into measurements
+     (user_id, value, unit, created_at, updated_at, captured_at,
+     measurement_import_id, location)
+     select #{self.user_id},cpm,'cpm', now(), now(), captured_at,
+     #{self.id}, computed_location
+     from bgeigie_logs WHERE
+     bgeigie_import_id = #{self.id}])
+ end
+
+ def import_measurements
```
matschaffer: Is this function used somewhere? Not seeing it in the PR

robouden: No, not used yet.. But worked for test on dev.

app/controllers/bgeigie_imports_controller.rb
```diff
@@ -34,6 +34,13 @@ def approve
     redirect_to @bgeigie_import
   end
 
+  def fixdrive
+    @bgeigie_import = BgeigieImport.find(params[:id])
```
evadne: looks like a potential attack factor as a scope is not used, same in other places?

app/models/bgeigie_import.rb
```diff
@@ -88,6 +88,24 @@ def finalize!
     self.update_column(:status, 'done')
   end
 
+  def fixdrive!
+    import_measurements_fix
+    update_counter_caches
+    confirm_status(:measurements_added)
+    self.update_column(:status, 'done')
```
evadne: 1. `self.` is not needed
2. might be advisable to wrap in transaction

app/controllers/bgeigie_imports_controller.rb
matschaffer: #approve is also unscoped. I also noticed we should have this action in the require_moderator list as well. Will push that change in a second.

app/models/bgeigie_import.rb
matschaffer: This format is consistent (for better or worse) with other methods in the class (finalize!, approve!)

seanbonner:
Bubbling this back up, as it was implemented and worked fine, but now seems buggy again. From admin pages there's no Fix Drive option until *after* a drive is approved, which is obviously problematic. Maybe something got moved accidentally?

app/views/bgeigie_imports/_head.html.erb
```diff
@@ -13,6 +13,14 @@
           <%- end -%>
         </li>
       <%- end -%>
+      <%- if !@bgeigie_import.awaiting_approval? && user_signed_in? && current_user.moderator? && !@bgeigie_import.done? -%>
```
matschaffer: @seanbonner definitely set to show only if drive is not awaiting approval (presumably approved). Should the condition be just visible to any moderator?

seanbonner: Yeah, we wouldn't approve a broken drive, so need to be able to try and "fix" it before approval.

nokton: I would say - show it always and remove any intelligence from when to show it - it’s useless after approval for sure

Pieter

robouden:
Sean,

The fix drive button was only implemented to get the measurements that had
MD5 issues in the measurements table, nothing else.

rob



Regards,
Rob Oudendijk Yuka Hayashi
http://yr-design.biz   http://oudendijk.biz  http://about.me/robouden
tel +81 80-22605966  Skype: robouden  Facebook:robouden
<http://on.fb.me/QeKw2P> twitter:robouden <http://bit.ly/RAiSTC>

Issue #143, opened 2013-10-07
wrong rendering on "null" values


thinrope:
Example: https://api.safecast.org/en-US/measurements/26027811
JSON is fine:
{
    "captured_at": null,
    "device_id": 35,
    "id": 26027811,
    "location_name": null,
    "original_id": 26027811,
    "unit": "cpm",
    "user_id": 6,
    "value": 18,
    "latitude": 35,
    "longitude": 135
}

However the HTML rendering is offset/borked:

![scr](https://f.cloud.github.com/assets/754020/1280844/74d16a46-2f59-11e3-974e-4450ac01d604.png)

matschaffer:
Fixed by #169

PR #169, opened 2015-04-27
#143 Wrong render on null value - fixed

anditto:
Fixed the rendering problem. The user_views_their_measurements wasn't running because it was lacking a _spec.rb as a the filename, so I moved that a valid file too.


anditto:
Having users confirm their email when signing up was breaking the tests, so I added a confirmed_at on the User Fabricator. There's other ways to fix it, but I think this is sufficient for now.

seanbonner:
@anditto This is failing the Travis CI build, can you check it out again and fix whatever the conflict is so then I can merge it?

matschaffer:
:+1: for this. All tests run on my machine.

As we discussed a bit in slack, the travis config is pretty old and likely not operable to travis at the moment.

matschaffer:
@seanbonner #174 should get travis running again if you want to get this PR tested.

matschaffer:
@anditto if you want to rebase this we can probably get a passing travis run on it now that #174 is merged: https://travis-ci.org/Safecast/safecastapi/builds/64197346

matschaffer:
Though as I mentioned in the chat, it might be good to find out if any of our envs contain/allow any cases where captured_at would be nil.

If they don't it might be better to change the test data and leave the vie code alone.

matschaffer:
Confirmed on dev. Merging

0_0cpm_35_2690 _136_1659 __span_class__translation_missing__title__translation_missing__en-us_measurements_show_measurements__measurements__span_____safecast_api

PR #174, opened 2015-05-26
Travis updates to get things running

matschaffer:
Uses postgres 9.2 since that's what prod uses. Also uses postgis extension as documented in http://docs.travis-ci.com/user/using-postgresql/#Using-PostGIS


matschaffer:
Note that failure is expected pending merge of #143 - the progress is that it's actually tests failing rather than setup erroring out

PR #176, opened 2015-05-27
Get tests passing

matschaffer:
This is an update to #169 which avoids changing any production code.


matschaffer:
Nm.... I now see that #169 was mainly aimed at fixing #143. I'll test it #169 on dev as-is and merge following confirmation.

PR #177, opened 2015-06-04
Some doc updates from info provided by Rob

### LINKED ISSUE #125: A general introduction to the apps here?
Hello @seanbonner!

Can you help me run the whole system in the local development environment? This sounds weird but I'm new to here. Before doing any actual work, can I get a overview picture of the whole system, not only this app but also the apps that is relying on this? I downloaded the code, made it run in my computer, created a user and a moderator, found some sample logs from the tickets here, successfully imported it, but what else can I do with the site?

[GeoSence](https://github.com/Safecast/GeoSense) is one of those calling this app, right? I also downloaded it, ran it. But I don't know what I can do with it, like that how can I connect this to my local safecastapi? 

When I tried to create a map from the index page, I got an error:

    TypeError: Invalid hex string:
        ...
        at /Users/eli/Projects/GeoSense/api/map/index.js:134:16
        ...

where `new Buffer(uuid.v4(), 'hex')` is called. I tired in the eclipse debugger, finding that `uuid.v4()` returns a string like "788f0bdc-a2a2-4b57-ac69-78a4a6d6ca31", which is not a hex. Is it a bug? 

Thanks!

matschaffer:
Fixes #125

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,30 @@
 # Safecast/safecastapi
 
-A ruby translation project managed on [Locale](http://www.localeapp.com/) that's open to all!
+The app that powers [api.safecast.org](https://api.safecast.org/)
 
-## Contributing to Safecast/safecastapi
+## Overview
 
-- Edit the translations directly on the [Safecast/safecastapi](http://www.localeapp.com/projects/public?search=Safecast/safecastapi) project on Locale.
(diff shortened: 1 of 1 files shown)


matschaffer:
@robouden Updated the readme a bit from your notes.

Issue #181, opened 2015-06-16
CircleCI is misconfigured


Steven-Wright:
CircleCI's database is not configured properly and fails on every PR.

Steven-Wright:
Needs to have it's own DB config, working on fix now.

@matschaffer, I'm not in the slack. Is there a reason why travisCI isn't fitting our needs, it seems to be configured and working?

PR #182, opened 2015-06-17
CI & Code Climate integration

Steven-Wright:
Configured CircleCI closing #181, configured Code Climate, and added master branch badges.

Note: if we ever start using CircleCI in parallel it will [break](http://docs.codeclimate.com/article/220-help-im-having-trouble-with-test-coverage#fyis) Code Climate coverage monitoring.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ rvm:
   - 1.9.3
 addons:
     postgresql: "9.2"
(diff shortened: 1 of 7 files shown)


matschaffer:
Fantastic! I'm fine with just one CI or the other but since they're both free I guessed can just run both for now.

PR #201, opened 2015-09-11
Rob changes

robouden:


db/schema.rb
```diff
@@ -181,6 +181,10 @@
     t.integer  "height"
     t.string   "surface"
     t.string   "radiation"
+    t.integer  "sensor_id"
```
matschaffer: Same case as before, if we don't have a migration here I'm almost definitely going to forget to add this to the prod db come deployment time. if you want to add them to prod manually ahead of time that should be fine.

evadne:
do we need foreign keys?

matschaffer:
Rejecting this now that we have Nick's proposed schema implemented in https://github.com/Safecast/safecastapi/tree/air/import

PR #210, opened 2015-10-16
Another couple of postgis-related changes in support of a staging env

matschaffer:
The db.rake part of this might not be necessary. Trying a new build of of staging now with an updated manifest from c66 support.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -2,9 +2,9 @@ namespace :db do
   #NOTE: Newer versions of rails should support this in schema.rb. Ours doesn't atm.
   desc 'Creates the postgis extension on the datatabase'
   task :enable_postgis => :environment do
-    # for some reason `create extension if exists` doesn't work with how the
-    # extension is normally getting added, so we'll look for the postgis schema
-    # instead
+    # for some reason `create extension if exists` doesn't work consistently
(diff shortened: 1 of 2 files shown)


spec/spec_helper.rb
```diff
@@ -1,6 +1,6 @@
 # Send test-coverage to code climate
 require "codeclimate-test-reporter"
```
evadne: Perhaps also invoke `require` only when there is a token

lib/tasks/db.rake
```diff
@@ -2,8 +2,19 @@ namespace :db do
   #NOTE: Newer versions of rails should support this in schema.rb. Ours doesn't atm.
   desc 'Creates the postgis extension on the datatabase'
   task :enable_postgis => :environment do
-    connection = ActiveRecord::Base.connection
-    connection.exec_query('CREATE EXTENSION IF NOT EXISTS postgis')
+    # for some reason `create extension if exists` doesn't work consistently
```
evadne: IMO this is unnecessary if we go with `structure.sql`

matschaffer: It is, that's why I closed the PR ;)

spec/spec_helper.rb
matschaffer: Good call. Opened as #215

PR #215, opened 2015-11-11
Wrap codeclimate init so we don't error if there's no token available

matschaffer:
This way we don't see warnings when running specs locally.


matschaffer:
Assigned to @evadne for review/merge

PR #233, opened 2016-02-09
Drive type 216

### LINKED ISSUE #216: Qualify drive type
add a meta data field for each drive. Drop down. Select - 1. Drive 2. Flight 3. Surface Default is Drive when not supplied by the API. For Drives approved, field should be set to “Drive” by script.

Right now there are a number flight/surface drives that are held up (can't be approved) since we can't effectively mix those with the drive (~1m) data.

robouden:
Fixes #216

### DIFF
--- a/db/migrate/20120417190549_add_details_to_measurement_import.rb
+++ b/db/migrate/20120417190549_add_details_to_measurement_import.rb
@@ -6,6 +6,8 @@ class AddDetailsToMeasurementImport < ActiveRecord::Migration
 
     add_column :measurement_imports, :orientation, :string
 
+    add_column :measurement_imports, :subtype, :string
+
     add_column :measurement_imports, :cities, :text
 
   end
--- a/db/migrate/20120417190549_add_details_to_measurement_import.rb
+++ b/db/migrate/20120417190549_add_details_to_measurement_import.rb
@@ -6,8 +6,6 @@ class AddDetailsToMeasurementImport < ActiveRecord::Migration
 
     add_column :measurement_imports, :orientation, :string
 
-    add_column :measurement_imports, :subtype, :string
-
     add_column :measurement_imports, :cities, :text
 
   end
--- a/db/migrate/20160208190731_add_subtype_to_measurement_imports.rb
+++ b/db/migrate/20160208190731_add_subtype_to_measurement_imports.rb
@@ -1,7 +1,21 @@
 class AddSubtypeToMeasurementImports < ActiveRecord::Migration
-  def change
-    add_column :measurement_imports, :subtype, :string, default: 'None', null: false
+  def up
+    execute <<-SQL
+      CREATE TYPE measurement_imports_subtype AS ENUM ('None', 'Drive', 'Surface', 'Cosmic');
+      ALTER TABLE measurement_imports
+        ADD COLUMN subtype measurement_imports_subtype
+        NOT NULL
+        DEFAULT 'None';
+    SQL
     add_index :measurement_imports, :subtype
     add_index :measurement_imports, [ :id, :subtype ]
+  end
+
+  def down
+    execute <<-SQL
+      ALTER TABLE measurement_imports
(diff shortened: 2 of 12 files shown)


db/migrate/20120417190549_add_details_to_measurement_import.rb
```diff
@@ -6,6 +6,8 @@ def change
 
     add_column :measurement_imports, :orientation, :string
 
+    add_column :measurement_imports, :subtype, :string
```
evadne: When changing the database one shall not add code to an old migration, because when deploying, Rails looks at the `schema_migrations` table to determine what migrations to run; consequently this new line will not have any effect on an existing production database waiting to be migrated.

The correct approach is to add another migration. Please refer to [Active Record Migrations](http://edgeguides.rubyonrails.org/active_record_migrations.html).

db/migrate/20160208190731_add_subtype_to_measurement_imports.rb
```diff
@@ -0,0 +1,7 @@
+class AddSyTypeToMeasurementImports < ActiveRecord::Migration
```
evadne: The name of this migration does not match the filename, therefore this migration will throw an error when executed.

db/migrate/20160208190731_add_subtype_to_measurement_imports.rb
```diff
@@ -0,0 +1,7 @@
...
+  def change
+
+  	add_column :measurement_imports, :subtype, :string
```
evadne: It looks to me that this column should not be nullable.

robouden: thanks

db/schema.rb
```diff
@@ -159,6 +159,7 @@
     t.text     "credits"
     t.decimal  "height",             :precision => 8, :scale => 2
     t.string   "orientation"
+    t.string   "subtype"
```
evadne: Considering `null: false`. Also, consider using [Postgres Enumerated Types](http://www.postgresql.org/docs/current/static/datatype-enum.html) to limit what this column can contain, which is more beneficial.

Frangible:
Sorry to add to complexity, but please review Evadne's 2016-02-09 comment re: enumerated types and non-NULLs.  Following Evadne's recommendations here will let us avoid significant recurring query costs and I/O overhead.  I believe it is worth the increased implementation complexity given the current DB loads.  

measurements is a very large table, and joins will have a cost, no matter the indices.  Controlling type width can have a significant impact.

Would suggest a default enumerated type of "none/unknown" or some NULL analog that will not have NULL's performance overhead.

eg, 
`measurement_import_subtypes`
`--|------------------------`
` 1 | None`
` 2 | Cosmic`
~~` 3 | Star Alignment`~~
~~` 4 | Elder Gods`~~
~~` 5 | Eldritch Knowledge`~~
~~` 6 | Madness`~~

Then, `measurement_import.subtype` becomes an integer foreign key reference to the enumerated type here as Evadne describes.  In the above example, that's `1` for existing data and `2` for the cosmic log imports.

Finally, don't forget to index it.  This pull request does not specify an index which means the server would probably die / timeout on the query and others due to the IOPS.  I would suggest an index on `measurement_import.subtype` and `measurement_import.subtype, `measurement_import.id`, following Evadne's suggestions above.

matschaffer:
@Frangible figure this will do the trick? Table looks like this on my laptop:

```
 created_at         | timestamp without time zone |
 updated_at         | timestamp without time zone |
 subtype            | character varying(255)      | not null default 'None'::character varying
Indexes:
```

Frangible:
As I've already said, subtype really needs to be either an enum, or an integer FK to a type table, or that join might really bottleneck things.

If we don't want to mess with `structure.sql` for now, the type table should only take like 3 lines of code and will save us a lot of potential headaches.  It's trivial to implement and has no other dependencies AFAIK.

PR #242, opened 2016-04-07
Added process button for unprocessed files

robouden:
Add button for forced processing

### DIFF
--- a/app/controllers/bgeigie_imports_controller.rb
+++ b/app/controllers/bgeigie_imports_controller.rb
@@ -27,7 +27,7 @@ class BgeigieImportsController < ApplicationController
   def new
     @bgeigie_import = BgeigieImport.new
   end
-  
+
   def approve
     @bgeigie_import = BgeigieImport.find(params[:id])
     @bgeigie_import.approve!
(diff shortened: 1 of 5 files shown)


robouden:
replaces 241

app/controllers/bgeigie_imports_controller.rb
```diff
@@ -39,6 +39,12 @@ def fixdrive
     @bgeigie_import.fixdrive!
     redirect_to @bgeigie_import
   end
+  
+  def process_button
+    @bgeigie_import = scope.find(params[:id])
+    @bgeigie_import.process_button!
```
matschaffer: Have you tried just calling `.process` here? Seems like the `.process_button!` method you've added is exactly the same as `.process`

app/models/bgeigie_import.rb
```diff
@@ -71,6 +71,14 @@ def process
     delete_tmp_file
   end
 
+def process_button!
```
matschaffer: If `.process` works, we can remove this method.

app/controllers/bgeigie_imports_controller.rb
robouden: Yes.. tried that first but I created errors.. I think i can fix that..

app/models/bgeigie_import.rb
robouden: Originally I tried.. But it did not work.. not sure why..

PR #244, opened 2016-04-16
Drive type 216 enums

matschaffer:
@Frangible This should do the trick on drive types as enums

Note: That it's dependent on #214 so probably best not to merge as-is just opening this PR for inspection.

Resulting table is:

```
 created_at         | timestamp without time zone | not null
 updated_at         | timestamp without time zone | not null
 subtype            | measurement_imports_subtype | default 'None'::measurement_imports_subtype
Indexes:
```


matschaffer:
@seanbonner @robouden @Frangible @evadne 

This seems to work fine. One gotcha to be aware of is that the admin panel doesn't seem to know how to edit/display the enum field.

Could be we need to grab a more recent version, but I don't see that as a blocker since we should be able to edit from the regular screen as a moderator anyway.

matschaffer:
Pushed this to origin:drive-type-216 so we can pick it up in #233

PR #253, opened 2016-04-29
Use -i option when executing pg_dump if its version is less than 9.0.0

eitoball:
This PR changes to use `-i` option in `pg_dump` command only when `pg_dump` version is less than 9.0 because

1. From version 9.5, `pg_dump` returns error ("invalid option").
2. This option ( `-i` or `--ignore-version`) is "A deprecated option that is now ignored." in `pg_dump` version 9.[0-4].

### DIFF
--- a/config/initializers/postgis.rb
+++ b/config/initializers/postgis.rb
@@ -13,7 +13,9 @@ require 'rake'
   username = config['username']
   filename = ::Rails.root.join('db', 'structure.sql')
   database = config['database']
-  `pg_dump -i -U "#{ username }" -s -x -O -f #{ filename } #{ database }`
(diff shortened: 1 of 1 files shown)


matschaffer:
Thanks @eitoball sorry I missed you opening this.

Are we sure we want to detected versions or should we just drop the `-i` usage entirely?

The prod db is 9.3 anyway. I think staging is 9.4.

eitoball:
> Are we sure we want to detected versions or should we just drop the -i usage entirely?
> 
> The prod db is 9.3 anyway. I think staging is 9.4.

If databases are 9.3 or above, then we should just drop `-i` usage. Updating commits...

matschaffer:
Thanks!

PR #263, opened 2016-06-05
Fix #262

### LINKED ISSUE #262: Cannot filter by subtype column in bGeigie import query.
**Summary**

Recently, a new column was added to measurement_imports - subtype.  However, this is not reflected in the API, making it impossible to query specifically for drives or cosmic radiation data.

(eg, the SQL: `WHERE subtype = 'Cosmic'`)

**Steps to Reproduce**

1. In a browser, go to: https://api.safecast.org/bgeigie_imports.json?order=created_at+asc&subtype=Cosmic
2. Verify the subtype field of the results set.

**Expected Behavior**

Only imports with subtype=Cosmic are returned.

**Actual Behavior**

All imports are returned.

**Other Notes**

This should also ideally support querying for multiple subtype values to be useful, as currently drive imports are split across the values "None" and "Drive".

For example: `https://api.safecast.org/bgeigie_imports.json?order=created_at+asc&subtype=Drive,None`

Or, in SQL: `WHERE subtype IN ('Drive', 'None')`

A possible fix(?) may be:

1. https://github.com/Safecast/safecastapi/blob/master/app/controllers/bgeigie_imports_controller.rb
 * Add: `has_scope :by_subtype`
2. https://github.com/Safecast/safecastapi/blob/master/app/models/bgeigie_import.rb#L36
 * `def self.by_subtype(subtype)`
 * `    where(:subtype => subtype)`
 * `end`

However, that does not address the issue of needing to filter by both "None" and "Drive" simultaneously to get all drive results as expected.

eitoball:
This PR should fix #262

### DIFF
--- a/app/models/bgeigie_import.rb
+++ b/app/models/bgeigie_import.rb
@@ -51,8 +51,8 @@ class BgeigieImport < MeasurementImport
   end
 
   class << self
-    def subtype(type)
-      where(subtype: type)
+    def by_subtype(type_or_types)
+      where(subtype: type_or_types)
     end
   end
 
--- a/app/models/bgeigie_import.rb
+++ b/app/models/bgeigie_import.rb
@@ -50,6 +50,12 @@ class BgeigieImport < MeasurementImport
     where("created_at > ?", time)
   end
 
+  class << self
+    def subtype(type)
+      where(subtype: type)
+    end
+  end
+
   def name
     read_attribute(:name).presence || "bGeigie Import ##{id}"
   end
--- a/app/models/bgeigie_import.rb
+++ b/app/models/bgeigie_import.rb
@@ -50,10 +50,8 @@ class BgeigieImport < MeasurementImport
     where("created_at > ?", time)
   end
 
-  class << self
-    def by_subtype(type_or_types)
-      where(subtype: type_or_types)
-    end
+  def self.by_subtype(type_or_types)
+    where(subtype: type_or_types)
   end
 
   def name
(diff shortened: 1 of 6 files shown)


app/models/bgeigie_import.rb
```diff
@@ -50,6 +50,12 @@ def self.uploaded_after(time)
     where("created_at > ?", time)
   end
 
+  class << self
```
matschaffer: Is there a reason for using `class << self` here instead of `def self.` like the other scoping methods? I'm find with either, but consistency is always good.

spec/fabricators/bgeigie_import_fabricator.rb
```diff
@@ -1,4 +1,4 @@
 Fabricator(:bgeigie_import) do
-  source File.new(Rails.root.join("spec/fixtures/bgeigie#{Fabricate.sequence}.log"))
+  source { sequence(:source) { |n| Rails.root.join("spec/fixtures/bgeigie#{n % 6}.log").open(File::RDONLY) } }
```
matschaffer: The `RDONLY` is interesting. Is it just for safety or was there something in the test that called for it?

app/models/bgeigie_import.rb
eitoball: You are right. I will rewrite it.

spec/fabricators/bgeigie_import_fabricator.rb
matschaffer: Gotcha. I tend to see a fair bit of `File.new(pathname_object)` but seems like either is basically the same.

As an aside I thought maybe `File.new` doesn't open file handlers, but looks like it does. TIL

PR #266, opened 2016-06-05
Refactor db:structure:{dump,load} rake tasks

eitoball:
This PR refactors db:structure:{dump,load} rake tasks because

* RGeo::ActiveRecord::TaskHacker will be removed in rgeo-activerecord 1.0.0 or above.
* Files in config/initializers will be counted for test coverage. It would be nice to exclude from coverage.

coverage_-_safecast_safecastapi_-_code_climate

CI seems to be passing, and it means 'db:structure:load` is working. As long as I tested, `db/structrue.sql' generated by db:structure:dump rake task seems to almost same.


matschaffer:
Looks good. I tested by also creating a bogus migration and the resulting sql showed only my added column. Seems to be working well. Thanks!

PR #281, opened 2016-06-09
Stop exporting measurments in csv

robouden:
kaminari counting tweak


matschaffer:
Interesting. Was there a slack conversation around this?

nokton:
Why is this requested?

 Pieter

eitoball:
No conversation nor request. I was going to propose to remove this feature. Mainly because csvbuilder gem has not been maintained and cannot be used in rails 4.

Maybe, we can look at access log to see if this feature is used or not.

seanbonner:
I'd add that CSV is probably the most useful way we export the measurements and I'm constantly hearing from people about how glad they are we do it, so I don't think it's something we should stop at all.

matschaffer:
FWIW, this doesn't impact the availability of https://api.safecast.org/system/measurements.tar.gz (the batch csv export)

The only match for `csv` in the logs is the test I just did, so doesn't seem like live CSV generation is really a used feature.

```
❯ ssh app1.safecast grep -i csv log/production.log
ActionView::MissingTemplate (Missing template bgeigie_imports/edit, application/edit with {:locale=>[:"en-US", :en], :formats=>[:html], :handlers=>[:erb, :builder, :jbuilder, :csvbuilder, :haml]}. Searched in:

~
❯ ssh app2.safecast grep -i csv log/production.log
[80c4a0068c252dc0e03c96cff89a2322] [162.243.100.197] Started GET "/measurements.csv" for 162.243.100.197 at 2016-06-10 00:40:06 +0000
[80c4a0068c252dc0e03c96cff89a2322] [162.243.100.197] Processing by MeasurementsController#index as CSV
[80c4a0068c252dc0e03c96cff89a2322] [162.243.100.197]   Rendered measurements/index.csv.csvbuilder (0.2ms)
[6b42d0b5fa750ffa2d792240171c5156] [162.243.100.197] Started GET "/measurements.csv" for 162.243.100.197 at 2016-06-10 00:40:11 +0000
[6b42d0b5fa750ffa2d792240171c5156] [162.243.100.197] Processing by MeasurementsController#index as CSV
[6b42d0b5fa750ffa2d792240171c5156] [162.243.100.197]   Rendered measurements/index.csv.csvbuilder (0.2ms)
```

thinrope:
As discussed on Slack this just broke the graphs on realtime.safecast.org ...
I think having a CSV is very important, so alternative ways to generate it might be needed.

matschaffer:
Should be online again momentarily now that #286 is merged

nokton:
Yep - I also was wondering why this got deleted - lets revert back plz

nokton:
We have far more pressing items in the backlog than these items - can we do some sorting of the items. Especially items requested by volunteers or projects should be on top of the list.

matschaffer:
@nokton it's already reverted. https://waffle.io/Safecast/Safecastapi has the "ready" list that we tend to work basically top-to-bottom. 

What should be toward the top that isn't already?

nokton:
Thanks!

PR #282, opened 2016-06-09
Revert "Not issue count query in kaminari"

eitoball:
Reverts Safecast/safecastapi#275 as mentioned in https://safecast.slack.com/archives/api/p1465468112000291

I will investigate [count estimate](https://wiki.postgresql.org/wiki/Count_estimate)


matschaffer:
Cool by me. Since number-of-measurements gets looked at a lot I'd really like to see a cached value + daily cron implementation at some point. That way we can display it prominently w/o much DB overhead.

nokton:
Do you have the SQL statement that is used to get the total row count on the measurement table handy by any chance?

matschaffer:
@nokton `SELECT COUNT(*) FROM "measurements"`

This PR is deployed btw so count is available at the top of https://api.safecast.org/en-US/measurements as well as at http://api.safecast.org/count

nokton:
Hmmm - you may want to try to get an estimated count instead:

https://wiki.postgresql.org/wiki/Count_estimate

BTW - Which version of Postgres we’re running?

matschaffer:
9.2.5 on production

nokton:
Will be good to see if the “estimate” works better - it’s OK for this number to be a “rough” number

matschaffer:
Looks like this _might_ be a little further off than we'd like:

```
safecast=# SELECT count(*) FROM measurements;
  count
----------
 49094666
(1 row)

safecast=# SELECT reltuples AS approximate_row_count FROM pg_class WHERE relname = 'measurements';
 approximate_row_count
-----------------------
           3.79935e+07
(1 row)

safecast=#  SELECT count_estimate('SELECT * FROM measurements');
 count_estimate
----------------
       51252988
(1 row)
```

PR #283, opened 2016-06-12
Fix some rubocop offences

eitoball:
This PR enables some of Rubocop check.


app/helpers/bgeigie_imports_helper.rb
```diff
@@ -2,20 +2,20 @@ module BgeigieImportsHelper
...
   end
 
-  def bgeigie_nav_li(status)
+  def bgeigie_nav_li(status) # rubocop:disable Metrics/AbcSize
     active = if params[:by_status].blank?
       status == :all
     else
       params[:by_status] == status.to_s
-    end
+             end
```
matschaffer: Weird alignment here. Seems like maybe the rest of the if block should get indented as well.

app/helpers/bgeigie_imports_helper.rb
```diff
@@ -2,20 +2,20 @@ module BgeigieImportsHelper
...
+      {
         :lat => b.latitude, 
         :lng => b.longitude,
         :cpm => b.cpm
       }
     end
   end
 
-  def bgeigie_nav_li(status)
+  def bgeigie_nav_li(status) # rubocop:disable Metrics/AbcSize
```
matschaffer: Looks like 9 counts of `Metrics/AbcSize` override in this PR alone.

Are we sure we don't want to increase the figured max?

config/initializers/secret_token.rb
```diff
@@ -4,4 +4,6 @@
 # If you change this key, all old signed cookies will become invalid!
 # Make sure the secret is at least 30 characters and all random,
 # no regular words or you'll be exposed to dictionary attacks.
+# rubocop:disable Metrics/LineLength
 Safecast::Application.config.secret_token = '39c00ee18c6a2edd472f8f8b5954b6a332636e32b4c80ed9abc9053cc33bf7f1913d4cb6267ea741d10b48acfd9b3b3df2336fd02de83a4290509bc5d97c0edd'
+# rubocop:enable Metrics/LineLength
```
matschaffer: Another case where it feels like maybe the default is too strict. Especially since I believe this is basically a line generated by rails.

app/controllers/application_controller.rb
```diff
@@ -30,19 +29,19 @@ def options
...
   
-  def rescue_action(env)
+  def rescue_action(_env)
     respond_to do |wants|
       wants.json { render :json => "Error", :status => 500 }
     end
   end
 
-  def cors_set_access_control_headers
+  def cors_set_access_control_headers # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
matschaffer: 10 counts of Metrics/MethodLength too. Seems like it previously detected 22 as the threshold. Maybe call it 20 and see if we can reduce the peppering of disables?:

matschaffer:
Seems like at least a handful of these disables could be avoided by tweaking the code. I'm okay to merge if the plan is to do that next. Just don't want to get people into habit of: commit -> fail build -> add rubocop exception

eitoball:
> Seems like at least a handful of these disables could be avoided by tweaking the code. I'm okay to merge if the plan is to do that next. Just don't want to get people into habit of: commit -> fail build -> add rubocop exception

First, I would like to enable all rubocop check , and I am leaving current code as is for now. We will refactor to remove those disables, and adjust rubocop configuration values (e.g. `Metrics/LineLength`).

config/initializers/secret_token.rb
eitoball: Ya, 80 is probably too strict. maybe, somewhere around 100 - 128. I will change the value once all rubocop checks are enabled.

app/helpers/bgeigie_imports_helper.rb
eitoball: I would rather decrease to like 10. I think that the size is currently 15, which is quite complex.

matschaffer:
Works for me then. Merging and look forward to the refactorings.

app/helpers/bgeigie_imports_helper.rb
eitoball: I believe that the indentation of `else` will be checked by another rubocop rule. I would like to leave it for now since rubocop doesn't warn (yet).

app/helpers/bgeigie_imports_helper.rb
robouden: I think 15 was needed. Please check with Robin.

rob



Regards,
Rob Oudendijk Yuka Hayashi
http://yr-design.biz   http://oudendijk.biz  http://about.me/robouden
tel +81 80-22605966  Skype: robouden  Facebook:robouden
<http://on.fb.me/QeKw2P> twitter:robouden <http://bit.ly/RAiSTC>

app/controllers/application_controller.rb
eitoball: Maybe, 20, we will see. BTW, Sandi Metz says `Five lines per method` ( [Sandi Metz' Rules For Developers](https://robots.thoughtbot.com/sandi-metz-rules-for-developers) )

PR #286, opened 2016-06-13
Revert "Stop exporting measurments in csv"

eitoball:
Reverts Safecast/safecastapi#281

Created this PR just in case.

I would like to hear about use case of csv export of measurements if it is needed.


matschaffer:
I'd like to hold on merging until we know how this was used. If removing it did indeed break something we should make sure we have some sort of insight into that breakage.

Issue #277, opened 2016-06-08
Add ability to download drive log file in KML format from the drive page (request from volunteers in Fukushima)


nokton:

matschaffer:
@nokton did the volunteers have any example documents we could use?

I'm sure we could do the transformations, but KML looks like a really flexible format so might be worth finding out more about just what the intended use is we can try to make sure our structure matches that.

eitoball:
I wrote a simple script that converts bGeigie Import file to KML to learn about bGeigie import data and KML. [This map](https://www.google.com/maps/d/viewer?mid=1CIqDsjgvhxgpKQgmyQmpA8-CirM) is created from [this bGeigie Import](https://api.safecast.org/en-US/bgeigie_imports/10775).

Maybe, we can discuss how we should format based on this?

robouden:
Eito-san

Nice. Did you have a look at the databot code at
https://github.com/Safecast/Databot?

regards
rob



Regards,
Rob Oudendijk Yuka Hayashi
http://yr-design.biz   http://oudendijk.biz  http://about.me/robouden
tel +81 80-22605966  Skype: robouden  Facebook:robouden
<http://on.fb.me/QeKw2P> twitter:robouden <http://bit.ly/RAiSTC>

nokton:
databot has the code to convert too - volunteers requested what databot does in the API (so when you hit download, you could have option to choose the format)

We could even include an option to download the PDF map of the drive like in databot

eitoball:
@robouden 

> Did you have a look at the databot code at

No, I didn't know about the project. I checked code, and updated scriptand [map](https://www.google.com/maps/d/viewer?mid=1CIqDsjgvhxgpKQgmyQmpA8-CirM) (except name in uSv/h) . I am going to start to work on this feature.

robouden:
Eito-san,

If you need more help let me know. I have also a standalone version that
Lionel made a few years ago.

regards
rob



Regards,
Rob Oudendijk Yuka Hayashi
http://yr-design.biz   http://oudendijk.biz  http://about.me/robouden
tel +81 80-22605966  Skype: robouden  Facebook:robouden
<http://on.fb.me/QeKw2P> twitter:robouden <http://bit.ly/RAiSTC>

nokton:
Databot is on Lionel's github repo: 
https://github.com/bidouilles/Databot

 Pieter

PR #287, opened 2016-06-14
Feature #277 - Download drive log file in KML format

eitoball:
See #277 for details.

You can download KML format file.

bgeigie_import__6____bgeigie_imports____safecast_api

The result of importing generated file.

map2_png

For comparison, this is a map on bgeigie import page (i.e. https://api.safecast.org/en-US/bgeigie_imports/10757)

map1_png


nokton:
That’s super - many volunteers will be happy! Great job!

app/controllers/bgeigie_imports_controller.rb
```diff
@@ -95,6 +95,15 @@ def destroy
     end
   end
 
+  def kml
+    bgeigie_import = BgeigieImport.find(params[:id])
+    ::Actions::BgeigieImports::Kml.new(
```
matschaffer: Do you have more info on this `Actions` concept? I'm confused why you'd go this route over letting ActionView process the ERB for you.

config/initializers/mime_types.rb
```diff
@@ -1,8 +1,7 @@
...
-# Mime::Type.register "text/richtext", :rtf
-# Mime::Type.register_alias "text/html", :iphone
-
-Mime::Type.register "application/vnd.safecast+json;version=1.0", :safecast_api_v1_json
-Mime::Type.register "application/vnd.safecast+xml;version=1.0",  :safecast_api_v1_xml
+{
+  safecast_api_v1_json: 'application/vnd.safecast+json;version=1.0',
+  safecast_api_v1_xml: 'application/vnd.safecast+xml;version=1.0',
+  kml: 'application/vnd.google-earth.kml+xml'
+}.each do |sym, str|
```
matschaffer: Feels a little YAGNI here. Do we have plans for adding many more mime types you think?

app/controllers/bgeigie_imports_controller.rb
eitoball: I was gonna use `render`, but it didn't work maybe because of extension (.kml). I will try again once we update to rails 4. I will add comment about it later.

For `Actions` concept, 2 reasons. 1) Mentioned above 2) ABC size.
 If we can do like `render 'beigie_logs', formats: [:kml], content_type: Mime::KML`, we can get rid of this class. Although, I like concept of "hexagonal architecture" or "use case" and would like to extract some of actions into classes for reduce class length and/or method complexity metrics.

config/initializers/mime_types.rb
eitoball: > Do we have plans for adding many more mime types you think?

Probably no. I am fine with either way.

matschaffer: Let's just go with the original version then.

app/controllers/bgeigie_imports_controller.rb
matschaffer: _that_ makes sense. `render 'beigie_logs', formats: [:kml], content_type: Mime::KML` is what I would have expected, but if it doesn't like the kml then I'm cool with this and we can move on.

Maybe add a comment or gh issue to revisit after rails 4?

config/initializers/mime_types.rb
eitoball: Done in 71c0672561de4c4d6b25108c1d7ef87b5ab538a1

app/controllers/bgeigie_imports_controller.rb
eitoball: Commented in 7a0685d607f1042f71a62f034eb99b58b460e9d1

PR #289, opened 2016-06-14
Add a comment about instance variables that are used in csvbuilder

eitoball:
I was surprised when I read document on https://github.com/gtd/csv_builder


matschaffer:
Good catch 👍

PR #298, opened 2016-06-22
Update unicorn to 5.1.0

eitoball:
I am not sure if it helps current Rails 4 issue, but we should try...


robouden:
I will give it a try.

sasharevzin:
@eitoball did it help? And what was the issue?

PR #299, opened 2016-06-22
Move ssl enforcement to controller so we can allow plain text json requests

matschaffer:
Source: https://github.com/department-of-veterans-affairs/caseflow-efolder/commit/c05730e5c0837c9c968ffdfac0fd83149a45e438


eitoball:
LGTM

Maybe we refactor to use string constants instead later for performance reason.

eitoball:
:shipit:

PR #302, opened 2016-06-22
Fix condition for force_ssl and enable

eitoball:
It seems that condition (`ssl_enabled?`) to enable SSL access always returns `false`. This condition is expected to return `false` only when coming request is in JSON format.

Currently, we have

```ruby
force_ssl if: :ssl_enabled?

def ssl_enabled?
  Rails.env.production? && !request.format.symbol == :json
end
```

but,

```pry
[1] pry(main)> !Mime::JSON.symbol == :json
=> false
[2] pry(main)> !Mime::HTML.symbol == :json
=> false
```

so, the `ssl_enabled?` always returns `false`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
 
   force_ssl if: :ssl_enabled?
 
+  before_filter :strict_transport_security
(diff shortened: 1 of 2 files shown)


matschaffer:
Nice. Thanks for the test!

PR #309, opened 2016-06-24
Improved awaiting approval email

jsidd:
Added Log Name to subject line
Added all details to body


nokton:
Great!

app/mailers/notifications.rb
```diff
@@ -21,7 +21,7 @@ def import_awaiting_approval(import)
     @import = import
     mail(
       :to => moderators, 
-      :subject => "A Safecast import is awaiting approval")
+      :subject => "A Safecast import is awaiting approval - " << import.filename)
```
eitoball: This is okay, but using string interpolation is more ruby-ish. For example, `"A Safecast ... approval - #{import.filename}"`

PR #310, opened 2016-06-25
Fix datetime format in JSON, not to include milliseconds

eitoball:
Rails 4 includes millisecond in datetime string of JSON format. This causes Drivecast iOS app an error described in Safecast/Drive-ios#4 .

### DIFF
--- a/config/initializers/datetime.rb
+++ b/config/initializers/datetime.rb
@@ -1,0 +1,7 @@
+[Date, DateTime, Time, ActiveSupport::TimeWithZone].each do |klass|
+  klass.class_eval do
+    def as_json(*)
+      xmlschema(0)
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


Frangible:
My opinion: we should consider keeping milliseconds.  The API is delivering an ISO/RFC complaint date.  Its behavior is correct.

iOS app: v1.1 is currently in testing and pending review by Apple.  It is passing all tests thus far.  It correctly handles dates from the API with or without the milliseconds, as per RFC/ISO standard.  (tested locally via http://api-staging.safecast.org vs. https://api.safecast.org)

However, we will need to support no-MS dates for a short period in the future until the app update is approved and users have a chance to download it.

matschaffer:
@eitoball @Frangible how about we add a version number or a date to that comment so we know when it's (reasonably) safe to put the milliseconds back?

Frangible:
v1.1 for the iOS app.  Date of Apple approval unknown.  7 days from this post seems likely safe.

eitoball:
> how about we add a version number or a date to that comment so we know when it's (reasonably) safe to put the milliseconds back?

How about Android version? Any official clients?

Frangible:
> How about Android version? Any official clients?  
  
Android app is "Safecast:Drive".  From what I saw on Slack, it appears unaffected.

matschaffer:
Closing this now that the iOS app has been updated.

PR #319, opened 2016-07-06
Initial work on docker/beanstalk deployment

matschaffer:
Not yet ready for merging.

Connected to #329


Dockerfile.base
```diff
@@ -0,0 +1,35 @@
+FROM ubuntu:16.04
+
+RUN apt-get update -y -qq
+RUN apt-get install -y -qq \
```
eitoball: Are `libyaml-dev`, `libffi-dev`, and `libgdbm-dev` included as dependency? If not, please install.

Dockerfile.base
```diff
@@ -0,0 +1,35 @@
...
+  zlib1g-dev \
+  libxml2-dev \
+  libxslt-dev \
+  libpq-dev \
+  nodejs \
+  postgresql-client
+
+ENV RUBY_VERSION 2.1.10
+
+RUN curl -O https://cache.ruby-lang.org/pub/ruby/2.1/ruby-${RUBY_VERSION}.tar.bz2 &&\
```
eitoball: I like using rbenv (ruby-build) or rvm to install ruby. They contain knowledge of  where to get ruby source and how to compile. Maybe, next time we have issues of installing ruby.

Gemfile
```diff
@@ -8,6 +8,7 @@ gem 'devise-token_authenticatable'
 gem 'cancan'
 
 gem 'rails', '4.0.13'
+gem 'puma'
 gem 'unicorn'
```
eitoball: If we use puma, can we remove unicorn gem?

matschaffer: Pretty sure not until we're clear of cloud66

Dockerfile.base
matschaffer: Rvm felt like overkill, but Ruby-build might be worth while.

Dockerfile.base
matschaffer: I didn't hit those as issues when getting the app running. Do you know if they're needed for something today? Easy enough to change the base container when needed.

eitoball: I thought that libyaml-dev is must for `YAML`. ubuntu 16.10 uses ruby 2.0 or above? If so, libyaml-dev might be already installed.

> Easy enough to change the base container when needed.

Okay. We might need to run specs in container. I will research Travis CI and Circle CI if they can run specs in docker container, and will try to set up Wercker (wercker.com)

matschaffer: Not installed, but yaml seems to parse fine at list via simple testing:

```
❯ docker exec -it safecastapi_web_1 /bin/bash
root@1ae4b05c7b2d:/app# dpkg -l libyaml-dev
dpkg-query: no packages found matching libyaml-dev
root@1ae4b05c7b2d:/app# irb
irb(main):001:0> require 'yaml'
=> true
irb(main):002:0> p YAML.load('a: b')
{"a"=>"b"}
=> {"a"=>"b"}
```

PR #322, opened 2016-07-11
Add pry-byebug

eitoball:
For debugging.

See https://github.com/deivid-rodriguez/pry-byebug


eitoball:
Merging since this should not affect production.

matschaffer:
👍

PR #328, opened 2016-09-02
Fix issue that users cannot reset their password

eitoball:
https://safecast.slack.com/archives/api/p1472746207000002


matschaffer:
I :heart: eito's attention to detail.

Issue #338, opened 2016-12-27
Update Ruby version


Morred:
At the moment we're using Ruby 2.1.10 (current stable is 2.4.0, I was able to update to 2.2.3 locally without having to fix/change anything). Any reasons why we should stick with 2.1.10?

eitoball:
Not particular reason. I was kind of waiting for AWS migration. Would you like to submit PR to update ruby to 2.2.x?

FYI, I was updating ruby and rails in turn like ruby 2.0 -> rails 4.0 -> ruby 2.1 -> rails 4.1 -> (ruby 2.2 ?)

Morred:
Yup, can do. Will also check which Rails version we can update to like this without running into immediate issues.
(Sorry for the late reply, was on vacation in Germany and travelled around a lot for friends and family stuff)

eitoball:
> Yup, can do. Will also check which Rails version we can update to like this without running into immediate issues.

Please talk to @matschaffer when you update ruby. About Rails, we should be update to 4.2 without any issues. We need to ruby 2.2 or above to Update to 5.0, and we need some modifications.

eitoball:
Closing since question has been answered.

PR #339, opened 2017-01-03
Rbenv config

db0sch:
I added this small dot file `.ruby-version` in order to make the project rbenv compatible (in addition to rvm. it shouldn't affect the initial config).
Wanted to keep it for my own config at first, but @matschaffer suggested that it could be useful for others (maybe new contributors who works with rbenv, and not rvm)

Saw in the issues that we planned to update Ruby version (issue #338) 
of course, this file must also be updated to mention the same ruby version as the one used in the project.


eitoball:
FYI, `.ruby-version` should work on rvm, too.  (https://rvm.io/workflow/projects#supported-files)

Morred:
Cool, I did that locally but much better to keep it in the repo too 👍

db0sch:
way better to support both rvm and rbenv. 
thanks for the rvm resource

PR #340, opened 2017-01-04
Read ruby version from .ruby-version

eitoball:
SSIA


db0sch:
That's way clever this way. Thanks @eitoball

Issue #276, opened 2016-06-08
View of "stuck" drives for moderators


matschaffer:
This pertains to #45 and #196 but we could use a good way to handle identify drives that need attention. For example:
- not processed (processing errors)
- not approved (wrong metadata)

Things to include on the list:
- submitter
- link to drive details
- peak CPM
- current approval status

Hoping this will be a good first step in clearing out the backlog of data.

seanbonner:
Unusual reading would be good, so if a log has 95% readings within a certain level and then a few 10000% higher would be good to have note for that.

nokton:
Or just showing the peak value will do - a bit easier to implement. 

 Pieter

eitoball:
From sync-meeting on Dec. 21st, 2016:

* Also include drives that are waiting for approval?
* Be able to approve/dis-approve many drive at a time
* Flag valid/invalid-ish drives

nokton:
Yes, also include drives waiting for approval. 

Bulk approval is fine but lets discuss how we want to do this (the idea is to have some auto approval logic for drives that all look fine based on a few criteria and allow those to be bulk or auto approved)

Another problem is we have a growing number of drives that don’t load (0 rows) we need to know how many rows failed because of duplicates, versus other reasons. For drives that are stuck, we need a reload button where we can force the drive to load again (and hopefully pass)

matschaffer:
Pretty sure the 0 rows is #330 but yes, probably does intersect with this ticket a bit. Might even be basically the same page in the app.

nokton:
Hi Mat - do we know what is causing this to happen - would suggest the next priority thing to tackle on the list

robouden:
Pieter and  Mat,

We should implement the same error checking and cleaning that databot does.
Basically implementing Lionels script into the ruby app.

regards,
rob



Regards,
Rob Oudendijk Yuka Hayashi
http://yr-design.biz   http://oudendijk.biz  http://about.me/robouden
tel +81 80-22605966  Skype: robouden  Facebook:robouden
<http://on.fb.me/QeKw2P> linkedin:robouden
<https://www.linkedin.com/in/roboudendijk>

PR #341, opened 2017-01-09
Initial implementation of "stuck" drive views for moderators

eitoball:
bgeigie_imports____safecast_api

For #276 

This is an initial implementation of "stuck" drive views. Need more works like wording and such.


matschaffer:
Gonna push this out. Doesn't seem like it'll break anything and it should be easier to gather feedback having it on prod.

Issue #197, opened 2015-09-02
Replace legacy API map visualization


Frangible:
## Todos:

- [x] Put the BVM init in a function and call that from the _map partial so we don't do that work on every page load
- [x] Switch up binds per [nick's comment](https://github.com/Safecast/safecastapi/pull/362#issuecomment-297747324)
- [x] See about adding a scale to the page using [this sprite](https://github.com/Safecast/Tilemap/blob/master/scales64_fullcolor_240x854.png)

## Original description

The current API legacy map visualization has the following issues:
1. The markers have poor visibility against the basemaps.
2. The markers are not z-indexed nor decuttered, making it difficult for mods to see anomalies.
3. The server often fails to generate the marker .js data for large drives, timing out.

A solution to this has already been developed, and merely requires being plugged in:
1. The bGeigie Log Viewer, a component of the [Tilemap](https://github.com/Safecast/Tilemap).
2. Fully client side, no other deps.  Resolves all client issues, is faster, requires no server resources.

**Changes required:**  
1. Existing map:  
   - On the API server, there is a job or process that generates marker data from logfiles for the legacy map.  This can be disabled/removed when the page code is migrated.  
   - On the legacy map HTML page, whatever current JS or includes are being used should be obviously removed.
   - On the legacy map HTML page, the layout should be changed to remove the static information pane from the left side of the screen.  
2. New map:  
   - Including the necessary .js files for the bGgeigie Log Viewer and its worker.  
3. Invoking:  
   - Instanting the main controller class and invoking the function to load a log file.  

**Requirements:**  
- **Files:** [bgeigie_viewer.js](https://github.com/Safecast/Tilemap/blob/master/bgeigie_viewer.js) and [bgeigie_viewer_worker.js](https://github.com/Safecast/Tilemap/blob/master/bgeigie_viewer_worker.js)
- **Library Dependencies:** None, except Google Maps API v3
- **Repo:** [Safecast Tilemap](https://github.com/Safecast/Tilemap) which also contains a conceptual overview.

**After including the .js files:**
1. **EITHER** a) No File Transfer UI:
   - The databind object passed to the controller class must be modified such that the databind for the UI is set to NULL. 
   - `binds.elementIds.bv_transferBar` should be set to NULL to prevent showing the file transfer progress UI.  
   - This is the simplest approach and should require no other configuration. _(untested)_  
2. **OR** b) File Transfer UI.  
[...]

PR #362, opened 2017-04-27
Integration of JS tilemap

matschaffer:
Connected to #197


Frangible:
>"This pull request quality could be better"

Ouch.  Anyway, only thing I'd say is you can replace the "binds" object you have in the BVM init with this (I fixed the issue):

For app/assets/javascripts/application/bgeigie_imports.js:  
```
var binds = { urls: { bv_worker_min: $map.data('bv-worker-min') } };
```

matschaffer:
@Frangible 😆 yeah seems like codacy has a stick up its butt there. I'll swap up the binds cause that looks much nicer.

I also realized I think this is attempting to load that block on every page cause of how asset complication works so I'll flip that up a little as well so we don't do a bunch of work just to have it say "no #map_canvas found"

Frangible:
>this is attempting to load that block on every page

Wow.  I noticed on production `api.safecast.org`:

* It's loading the Google Maps API with every page, whether or not it even uses a map.  
* The old map implementation seems to be in some coffee file that gets included (but not minified) into `application.js` -- this too gets included on every single page.
* I don't see where the other files in `production.rb` are in the repo or are ever getting used(?). (`dobro.css worldmap.js prettify.css prettify/prettify.js api_docs.js`)

More PR-related stuff I noticed:

* `html5shiv.js` should probably be named `html5shim.js`  (a shiv is a tool for helping one make friends in prison)
* Don't include the worker JS file like a normal HTML script (`<script src=...>`) because JS workers use functions undefined in normal JS (self vs. this, import, etc) and it will likely generate runtime errors.
    * (loading the worker source is handled automatically by JavaScript when instantiating `new Worker("w.js")`)

>codacy has a stick up its butt there

[tfw seeing it criticize C-style naming conventions](http://i.imgur.com/iIpkTVp.png)

PR #366, opened 2017-06-29
Revise marker infowindow

Frangible:
Visual/auto CPM/uSv/h and log CPM/cp5s separated.  Also add secondary
infowindow view with summary stats for logfile.


Frangible:
https://api.safecast.org/en-US/bgeigie_imports/10860 would be a good test case becasue of the bad points

Issue #375, opened 2017-07-31
Upgrade Ruby version


charlesgreen:
Ruby version 2.1.0 is at end of life. Only critical security updates will be released. 
I tested with upgrading Ruby to version 2.3.4p301 to match [Beanstalk - Ruby version](http://docs.aws.amazon.com/elasticbeanstalk/latest/dg/concepts.platforms.html#concepts.platforms.ruby), however, there seems to be a breaking change that impacts the json and simplecov gems. 

Make sure that `gem install json -v '1.8.1'` succeeds before bundling.

In Gemfile:
  codeclimate-test-reporter was resolved to 0.5.2, which depends on
    simplecov was resolved to 0.11.2, which depends on
      json

matschaffer:
Sounds good. Let's just make sure the `Gemfile*` are updated accordingly. Hopefully it'll still just be a `bundle install` for folks to get up and running on the new version.

charlesgreen:
The change to Ruby should still be just a bundle install however they too would need to install the updated version of Ruby. I checked with rbenv and found that version `2.3.4p301` was available. We would still need to check with rvm to confirm the same build is available.

matschaffer:
@charlesgreen looks like rvm knows "2.3.4" but not "2.3.4p301" by that name. But it installed p301

```
~
❯ rvm use 2.3.4
Using /Users/mat/.rvm/gems/ruby-2.3.4

~
❯ ruby --version
ruby 2.3.4p301 (2017-03-30 revision 58214) [x86_64-darwin16]
```

So pretty sure we should be able to keep everything happy if we lock to 2.3.4 and let the running system determine which p-number to use.

eitoball:
[FYI] ruby versioning system has been changed to semantic versioning since 2.1, but they use the same build system or something. So, they still use patch version like 'p301'. `2.3.4` is like`2.3.0-p301`

matschaffer:
We'll keep working on this over time I'm sure, but for now c66 is deployed with 2.2.7. Since this ticket starts by referencing 2.1 I'll close and we can open a new ticket (or just a PR) for the 2.2 -> 2.3 move when the time comes.

PR #376, opened 2017-08-01
Update ruby to 2.2.7

eitoball:
Related to #375. The issue is to upgrade to 2.3.x, but I think that our Rails is little bit old for 2.3. I think that we can update ruby safely if we update ruby to 2.2 and then rails to 4.2.


eitoball:
This PR is not approved, but I assume it is since #377 is approved.

matschaffer:
nice. Good eye @eitoball thanks

PR #399, opened 2017-09-17
Add support for navigate away warning if data entry not complete

Frangible:


Frangible:
@matschaffer - I'm not sure why the Travis CI builds are failing.  This only changes a single JavaScript file I tested locally successfully.  But it still needs to be deployed etc normally so if the build is failing it's still bad even though it has a very limited scope that affects nothing else.

PR #402, opened 2017-10-19
Use environment variables for settings

eitoball:
This PR is for #401 

You can set environment variables or via `.env` file by using dotenv gem. There is a file called `.env.sample` for what variables you can configure.

```
# AWS access key for SES and S3
AWS_ACCESS_KEY_ID=
AWS_SECRET_ACCESS_KEY=

# S3 bucket name to store bGeigie logs
S3_BUCKET=

# Intercom
INTERCOM_APP_ID=
INTERCOM_API_SECRET=
INTERCOM_API_KEY=
```


eitoball:
> Have you checked the db dump to see if there are any other configurable we need to make env vars for?

No, I have not. But, available values should be defined in `config/configurable.yml`, and all values in the file have been replaced.

matschaffer:
Ah cool. Well I checked the db anyway and we're 👍

Issue #406, opened 2017-12-03
Remove "rejected" logs from "waiting to be approved" and "processed" tab


seanbonner:
On the log submission pages, drives that are rejected are still showing up on "waiting to be approved" counts on the main page and on "processed" pages which is confusing for moderators. Can we please Add a "rejected" tab that lists those, and then filter out anything that has been rejected from anything suggesting it's awaiting action?

matschaffer:
Closing in favor of https://github.com/Safecast/safecastapi/issues/406 since I've added a bit more info there.

PR #408, opened 2017-12-05
Exclude rejected drives from "Imports Awaiting Approval" count on Dashboard

eitoball:
This is part of #406. This PR should exclude to count rejected drive from "Imports Awaiting Approval" on Dashboard.

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -6,7 +6,4 @@ class DashboardsController < ApplicationController
     end
     @unapproved_bgeigie_imports = BgeigieImport.unapproved
   end
-
-  alias_method :new, :show
-  alias_method :index, :show
 end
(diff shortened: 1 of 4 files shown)


matschaffer:
@Lugulbanda he's referring the change on https://github.com/Safecast/safecastapi/pull/408/files#diff-21497849d8f00507c9c8dcaf6288b136L13

Since `dashboard` was defined as a resource it gets a bunch of extra routes (URLs) that then were getting taken back out again on https://github.com/Safecast/safecastapi/pull/408/files#diff-e4663ec6bcde373fc1ee998eb69725b1L10

By specifying `only: %i(show)` you only get one route rather than all 7 as shown on http://guides.rubyonrails.org/routing.html#crud-verbs-and-actions

matschaffer:
Maybe worth mentioning the value here is that it's cleaner, probably provides a bit of a speed boost, and avoids potential abuse of the URLs we never needed in the first place.

PR #415, opened 2018-01-10
Upgrade fog & carrier wave

matschaffer:
Upgrade from dependabot was a bit too naive and caused failures in prod. This seems to be a better approach given local testing.

If you want to test locally, set your `.env` like this:

```
AWS_ACCESS_KEY_ID=YOUR_SAFECAST_KEY_ID
AWS_SECRET_ACCESS_KEY=YOUR_SAFECAST_KEY_SECRET

S3_BUCKET=safecastapi-imports-dev-ap-northeast-1
```

The carrierwave initializer will attempt to use your `$USER` as a prefix, so take note incase there's something in your user name that is invalid as an s3 path.


matschaffer:
See https://github.com/Safecast/infrastructure/pull/11 (closed repo) for bucket creation details

matschaffer:
I realize now that we could probably fix master by just setting region to `us-west-2`, but at least this allows for testing of s3 uploads in both dev & staging.

Issue #179, opened 2015-06-15
Rotate SSL keys for api.safecast.org


atalax:
There is an SSL key in `doc/ssl/safecast.key` which seems to be the one used by http://api.safecast.org. There is also some postgres password in `doc/.pgpass`. This is probably not intentional.

Steven-Wright:
This is indeed the case:
I downloaded the live API cert and compared it against the key, they have the same public exponent and modulus.

```
steven$ echo -n | openssl s_client -connect api.safecast.org:443 > safecast.crt
steven$ openssl rsa -noout -modulus -in safecast.key | openssl md5
d56eeea688ffb168b7d86be1e01b2a72`
steven$ openssl x509 -noout -modulus -in safecast.crt | openssl md5      
d56eeea688ffb168b7d86be1e01b2a72
```
`doc/.pgpass` gets copied into place by `/deploy/after_restart.rb`.

Both of these are pressing security issues. I'm not sure exactly how the app is deployed from the repo but I see two easy ways out. 
1. move `doc` to a local directory on the server and provide a symlink in the repo
2. move `safecast.key` and `.pgpass` to a local directory, ignore them in git, and copy them in with a post-merge hook (this way when you pull the latest stable version it will copy them over)

```
#!/bin/sh
#!/usr/bin/env sh
PRIVATE=~/private
DOC_PATH=$PWD/doc

#copy the ssl directory into place
cp -R $PRIVATE/ssl $DOC_PATH
#copy the pgpass into place
cp $PRIVATE/.pgpass $DOC_PATH
```

I'd imagine that something similar is already happening to handle the files present in `.gitignore`.

robouden:
Working to get the keys removed and made a private repro for the keys.

Steven-Wright:
@robouden - if you are making a private repo perhaps a submodule is the best bet.

diffalot:
I would advise against using a submodule for keys, I think it's better to .gitignore them.

The way submodules track commit hashes, `git submodules update` would throw errors in any environment except production (assuming that only production has access to the keys repo).

Steven-Wright:
@diffalot, then as per your [article](https://diff.mx/2015/01/git-subtree/) maybe git-subtree is more appropriate or maybe it could be as simple as a git stash. As it stands now, how does this repo get deployed?

diffalot:
@Steven-Wright Asking how the repo gets deployed is the right question here.

[...]

PR #420, opened 2018-01-23
Remove old SSL files

matschaffer:
Fixes https://github.com/Safecast/safecastapi/issues/179


eitoball:
@matschaffer Did you use `git rm`-ed with `--cache` option?

matschaffer:
No, we don’t really need these since they’ve been exposed already.

And we can get them out of git history if we ever need to confirm any details.

Issue #221, opened 2015-12-08
Remove rsync between app nodes during API data export jobs


matschaffer:
# Description

We have a number of API export jobs that run periodically.

![screen shot 2018-01-17 at 11 36 04](https://user-images.githubusercontent.com/690/35022870-797015d2-fb7b-11e7-882c-5461b2385876.png)

They correspond to:
- https://github.com/Safecast/safecastapi/blob/master/script/dump_clean
- https://github.com/Safecast/safecastapi/blob/master/script/dump_measurements
- https://github.com/Safecast/safecastapi/blob/master/script/dump_measurements_ios

As the last step, they all attempt to rsync the exported data to the other server. This makes server replacement require a bit of finagling which is not fun and won't be conducive to an AWS autoscaling setup.

Would be better to make these jobs export to S3 and have the app redirect the published URLs (`/system/(file name)`) to a public s3 object.

**Note:** while we're working on these scripts, we should probably add a `set -eo pipefail` to them. Currently there are cases (e.g., sqlite3 not being installed) where the export will fail but the job will still "succeed" as far as the exit code is concerned and we won't know about the failure until someone reports map problems.

## Background from issue when originally opened

I disabled this and got this worker error from dump_measurements_ios:

```
ssh: connect to host lion.api.safecast.org.c66.me port 22: Connection timed out 
rsync: connection unexpectedly closed (0 bytes received so far) [sender] 
rsync error: unexplained error (code 255) at io.c(226) [sender=3.1.0]
```

https://github.com/Safecast/safecastapi/blob/master/script/dump_measurements_ios#L66 seems to be a step to ensure the resulting file can be served from both app servers.

Seems like something we should look to push to s3 instead.

matschaffer:
Ah. I think I might see what's happening. https://github.com/Safecast/safecastapi/blob/master/script/dump_measurements_ios#L66 seems to be a step to ensure the resulting file can be served from both app servers.

Seems like something we should look to push to s3 instead.

thinrope:
I can probably work on that, but will need guidance...
My thoughts are that we can safely drop those, stuff an old copy of measurements.tar.gz in the new server and work from there. The data is only occasionally downloaded, so a few days delay is fine.
Basically, if that is the last thing holding the migration, just ignore it and let's work on in a new issue like "Provide snapshots of data for download". Once the dB is migrated, I'll work directly with it to achieve similar output.

thinrope:
[...]

PR #423, opened 2018-01-27
Improve readability of the script

thinrope:
Improved the readability of the script, added variables, comments. Plan to unify the other scripts.
No changed functionality (yet).

This was tested in AWS partially, I guess the only way is to deploy and see if it works in prod.

Connected to https://github.com/Safecast/safecastapi/issues/221

### DIFF
--- a/script/dump_measurements
+++ b/script/dump_measurements
@@ -1,12 +1,36 @@
-#!/bin/sh
-psql -U safecast --host $POSTGRESQL_ADDRESS safecast -c "\copy (select captured_at, ST_Y(location::geometry), ST_X(location::geometry), value, unit, location_name, device_id, md5sum, height, surface, radiation, created_at, measurement_import_id from measurements order by created_at desc) to '/tmp/measurements.csv' csv;"
-echo "Captured Time,Latitude,Longitude,Value,Unit,Location Name,Device ID,MD5Sum,Height,Surface,Radiation,Uploaded Time,Loader ID" | cat - /tmp/measurements.csv > /var/deploy/api.safecast.org/web_head/shared/system/measurements-out.csv
-cd /var/deploy/api.safecast.org/web_head/shared/system && tar -czf measurements-out.tar.gz measurements-out.csv
-mv measurements-out.tar.gz measurements.tar.gz && mv measurements-out.csv measurements.csv
+#!/bin/bash
 
-# rsync to the other server (this only runs on kangaroo)
-rsync -Havq --bwlimit 10000 /var/deploy/api.safecast.org/web_head/shared/system/measurements.* reindeer.api.safecast.org.c66.me:/var/deploy/api.safecast.org/web_head/shared/system/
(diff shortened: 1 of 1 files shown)


script/dump_measurements
```diff
@@ -1,12 +1,36 @@
...
-# rsync to the other server (this only runs on kangaroo)
-rsync -Havq --bwlimit 10000 /var/deploy/api.safecast.org/web_head/shared/system/measurements.* reindeer.api.safecast.org.c66.me:/var/deploy/api.safecast.org/web_head/shared/system/
-
-#
-# Now this is available as https://api.safecast.org/system/measurements.tar.gz
-#
+#!/bin/bash
+
+# 0: configuration -------------------------------------------------
+THIS_SERVER="kangaroo"	# confirm!
```
matschaffer: `meerkat.api.safecast.org.c66.me` if you want to be thorough about it.

script/dump_measurements
```diff
@@ -1,12 +1,36 @@
...
+OTHER_SERVERS="reindeer.api.safecast.org.c66.me"	# space separated list
```
matschaffer: Nice to call out the behavior. 🤞  that we won't ever need another host on this list 😉

script/dump_measurements
```diff
@@ -1,12 +1,36 @@
...
+FINAL_DIR="/var/deploy/api.safecast.org/web_head/shared/system"
+FILENAME_STEM="measurements"
+
+DB_HOST="${POSTGRESQL_ADDRESS}"	# confirm, where is it set up in env?
```
matschaffer: c66 injects this into the configs which start the process

script/dump_measurements
```diff
@@ -1,12 +1,36 @@
...
+DB_USER="safecast"
+DB_DATABASE="safecast"
+DB_OUTFILE="/tmp/${FILENAME_STEM}.csv"
+# -----------------------------------------------------------------
+
+# 1: dump the data in csv format, no header
+psql --host "${DB_HOST}" --username "${DB_USER}" --dbname="${DB_DATABASE}" -c "\\copy (select captured_at, ST_Y(location::geometry), ST_X(location::geometry), value, unit, location_name, device_id, md5sum, height, surface, radiation, created_at, measurement_import_id from measurements order by created_at desc) to '${DB_OUTFILE}' csv;"
+
+# 2: add header and remove ${DB_OUTFILE}
+echo "Captured Time,Latitude,Longitude,Value,Unit,Location Name,Device ID,MD5Sum,Height,Surface,Radiation,Uploaded Time,Loader ID" | cat - ${DB_OUTFILE} > ${FINAL_DIR}/measurements-out.csv
```
matschaffer: Guess we're not trying to change behavior here, but curious why we'd generate our own headers rather than embedding them in the select statement.

thinrope: Yes, exactly. I'll sync with the other scripts, then change the SQL to include column names.

PR #425, opened 2018-01-31
WIP - Update to structure.sql from prod `rake db:structure:dump`

matschaffer:


matschaffer:
@eitoball I sort of wonder if https://github.com/Safecast/safecastapi/blob/master/lib/tasks/db.rake#L16 might be causing some trouble here. Looks like all the postgis functions have wound up in the dump as well.

eitoball:
> I sort of wonder if https://github.com/Safecast/safecastapi/blob/master/lib/tasks/db.rake#L16 might be causing some trouble here. Looks like all the postgis functions have wound up in the dump as well.

@matschaffer As you mentioned somewhere (on Slack?), I think that it is because Postgis extension was created in `public` schema. There are not much difference other than Postgis-related definitions are added in new `structure.sql`. And, if you see line like [here](https://github.com/Safecast/safecastapi/pull/425/files#diff-f9ae8314082d8229b38da27dd6c2642dR10022), you can see that new `structure.sql` doesn't prefix `postgis` to Postgis-related column type.

The reason we are trying to update is that you want to check indexes that are not in migration files exists or not. It seems that such indexes like [in current structure.sql](https://github.com/Safecast/safecastapi/pull/425/files#diff-f9ae8314082d8229b38da27dd6c2642dR10022) and [in structure.sql on rails 4.2 pr](https://github.com/Safecast/safecastapi/pull/422/files#diff-f9ae8314082d8229b38da27dd6c2642dR792). These index names are different, but create index on the same column. It must be naming rule have been changed in Rails 3.x and Rails 4.2.

matschaffer:
Closing for now since the dump has too much postgis junk. Will revisit once we're on newer versions of postgres/postgis.

PR #438, opened 2018-03-16
Some tweaks that should fix up repo badges

matschaffer:


matschaffer:
got 👌 from @Frangible so mergings away!

Issue #446, opened 2018-04-13
The "Uploaded Before/After" API filter seems not to work


juhele:
I tried to search how many LOG files did we upload after 2017-08-16 but the filter always ends with error "Something went wrong..."

Here is the link for my filtered search:
https://api.safecast.org/en-US/bgeigie_imports?utf8=✓&q=&approved=&by_user_id=671&uploaded_after=16%2F08%2F2017+09%3A00%3A00&uploaded_before=&by_status=&commit=Filter

Tried this both in Firefox and Vivaldi - both in Windows 7 and Kubuntu linux and this seems to be a bug in the API.

PR #447, opened 2018-04-14
Format date/time in ISO format in filter dialog

eitoball:
Error in #446 seems to be caused because filter dialog creates date/time in MDY format (like "14/04/2018 12:34:56"). This PR change filter to create in ISO format (like "2018/04/14 12:34:56"). It has used MDY format since beginning. I wonder if it was working before...


matschaffer:
Yeah, filter doesn't seem to work today on prod anyway. Thanks for the fix!

PR #460, opened 2018-06-04
Add "Waiting for clarification" status

eitoball:
This PR should add "Waiting for clarification" status (`waiting_for`) to bGeigie import model.

A bGeigie import becomes to `waiting_for` status when moderator sends e-mails to uploader, and it becomes to the original status when moderator "resolve"s it.

----

### "Send Email To Uploader" tab in bGeigie import detail page

screenshot1

If you send e-mail to uploader, import status becomes to "waiting for".

### bGeigie import detail page when import status becomes "waiting for"

screenshot2

### Sidebar navigation for moderators

sceenshot3

The imports that are in "waiting for" status won't be shown in "Not processed" nor "Not approved".

### "Contact Detail" tab in bGeigie import detail page

screenshot4

Connected to https://github.com/Safecast/safecastapi/issues/456

### DIFF
--- a/app/controllers/bgeigie_imports/waiting_for_controller.rb
+++ b/app/controllers/bgeigie_imports/waiting_for_controller.rb
@@ -1,0 +1,11 @@
+module BgeigieImports
+  class WaitingForController < ApplicationController
+    before_filter :authenticate_user!
+    before_filter :require_moderator
+
+    def index
+      @bgeigie_imports = BgeigieImport.where(status: :waiting_for).page(params[:page])
+      respond_with @bgeigie_imports
+    end
+  end
+end
--- a/app/controllers/bgeigie_imports_controller.rb
+++ b/app/controllers/bgeigie_imports_controller.rb
@@ -49,7 +49,7 @@ class BgeigieImportsController < ApplicationController # rubocop:disable Metrics
       administrator: current_user, previous_status: @bgeigie_import.status
     )
     # use update_columns to avoid validations of presence of cities and credits
-    @bgeigie_import.update_columns(status: :waiting_for)
+    @bgeigie_import.update_columns(status: :awaiting_response)
     redirect_to @bgeigie_import
   end
 
--- a/app/controllers/bgeigie_imports_controller.rb
+++ b/app/controllers/bgeigie_imports_controller.rb
@@ -2,7 +2,7 @@ class BgeigieImportsController < ApplicationController # rubocop:disable Metrics
   respond_to :html, :json
 
   before_filter :authenticate_user!, only: [:new, :create, :edit, :update, :destroy]
-  before_filter :require_moderator, only: [:approve, :fixdrive]
+  before_filter :require_moderator, only: [:approve, :fixdrive, :send_email, :resolve]
 
   has_scope :by_status
   has_scope :by_user_id
@@ -45,6 +45,11 @@ class BgeigieImportsController < ApplicationController # rubocop:disable Metrics
   def send_email
     @bgeigie_import = BgeigieImport.find(params[:id])
     @bgeigie_import.send_email(params[:email_body])
+    @bgeigie_import.uploader_contact_histories.create!(
+      administrator: current_user, previous_status: @bgeigie_import.status
+    )
+    # use update_columns to avoid validations of presence of cities and credits
+    @bgeigie_import.update_columns(status: :waiting_for)
     redirect_to @bgeigie_import
   end
 
@@ -120,6 +125,13 @@ class BgeigieImportsController < ApplicationController # rubocop:disable Metrics
     ).execute(self, bgeigie_import.source.filename + '.kml')
   rescue ActiveRecord::RecordNotFound
     render text: '404 Not Found', status: :not_found
+  end
+
+  def resolve
+    @bgeigie_import = scope.find(params[:id])
+    history = @bgeigie_import.uploader_contact_histories.last
+    @bgeigie_import.update_columns(status: history.previous_status) if history
+    redirect_to @bgeigie_import
   end
 
   private
--- a/app/helpers/bgeigie_imports_helper.rb
+++ b/app/helpers/bgeigie_imports_helper.rb
@@ -27,4 +27,10 @@ module BgeigieImportsHelper
       a << t(".#{k}") if v
     end
   end
+
+  def operation_button(bgeigie_import, action, text = t(format('.%s', action)))
+    form_for bgeigie_import, url: { action: action } do |f|
+      f.submit text, class: 'btn btn-primary'
+    end
+  end
 end
--- a/app/views/bgeigie_imports/_head.html.erb
+++ b/app/views/bgeigie_imports/_head.html.erb
@@ -46,7 +46,7 @@
                     <%= f.submit t('.process_button'), class: 'btn btn-primary' %>
                   <%- end -%>
                 </li>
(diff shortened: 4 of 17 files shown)


app/controllers/bgeigie_imports/not_processed_controller.rb
```diff
@@ -4,7 +4,7 @@ class NotProcessedController < ApplicationController
     before_filter :require_moderator
 
     def index
-      @bgeigie_imports = BgeigieImport.submitted.unapproved.page(params[:page])
+      @bgeigie_imports = BgeigieImport.where(status: :unprocessed).unapproved.page(params[:page])
```
matschaffer: A method helper like https://github.com/Safecast/safecastapi/blob/master/app/models/bgeigie_import.rb#L23 could be nice here.

app/controllers/bgeigie_imports/waiting_for_controller.rb
```diff
@@ -0,0 +1,11 @@
+module BgeigieImports
+  class WaitingForController < ApplicationController
+    before_filter :authenticate_user!
+    before_filter :require_moderator
+
+    def index
+      @bgeigie_imports = BgeigieImport.where(status: :waiting_for).page(params[:page])
```
matschaffer: `awaiting_response` might be a better word choice than `waiting_for`

app/controllers/bgeigie_imports_controller.rb
```diff
@@ -45,6 +45,11 @@ def reject
   def send_email
     @bgeigie_import = BgeigieImport.find(params[:id])
     @bgeigie_import.send_email(params[:email_body])
+    @bgeigie_import.uploader_contact_histories.create!(
+      administrator: current_user, previous_status: @bgeigie_import.status
+    )
+    # use update_columns to avoid validations of presence of cities and credits
```
matschaffer: Not for this PR, but I think these validations should get removed anyway and instead moved into controller-level check during the "submit for approval" process.

app/views/bgeigie_imports/_head.html.erb
```diff
@@ -5,33 +5,69 @@
...
+                <%= form_for [:fixdrive, @bgeigie_import] do |f| %>
+                  <%= f.submit t('.fixdrive'), class: 'btn btn-primary' %>
+                <%- end -%>
+              </li>
+            <%- else %>
+              <%- case @bgeigie_import.status -%>
+              <%- when 'processed' -%>
+                <%# when import has been processed, moderator could reject %>
+                <li>
+                  <%= form_for @bgeigie_import, url: { action: 'reject' } do |f| %>
```
matschaffer: Nice improvements here. Got me wondering if we should have a button helper as well rather than all these little `form_for` blocks.

spec/controllers/bgeigie_imports/waiting_for_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  end
+
+  describe 'GET #index' do
+    before do
+      get :index
+    end
+
+    it { expect(response).to be_ok }
+    # Cannot use .to be_present because AR::Relation that is empty
+    # returns true to #blank?
```
matschaffer: 🤦‍♂️

app/controllers/bgeigie_imports/waiting_for_controller.rb
eitoball: oh... thank you. renamed including controller name and such

app/controllers/bgeigie_imports/not_processed_controller.rb
eitoball: this change was not for this PR, so reverted

app/views/bgeigie_imports/_head.html.erb
eitoball: indeed. created helper

app/helpers/bgeigie_imports_helper.rb
```diff
@@ -27,4 +27,10 @@ def status_details(bgeigie_import)
       a << t(".#{k}") if v
     end
   end
+
+  def operation_button(bgeigie_import, action, text = t(".#{action}"))
```
matschaffer: Looks rubocop flagged the use of `#` on a `def` line here https://travis-ci.org/Safecast/safecastapi/builds/388262191?utm_source=github_status

Might probably read better to default nil then `||=` on the first function line anyway.

eitoball: It seems to be rubocop bug. It is string interpolation, but rubocop thinks that it is a comment... Use `format` instead.

matschaffer:
Pretty sure we can't test this in staging because of the `Rejected by: Password for user safecast: psql: fe_sendauth: no password supplied` error we get attempting processing. So will test when it gets to prod

matschaffer:
@eitoball I tried testing this on https://api.safecast.org/en-US/bgeigie_imports/32161 but the drive doesn't seem to show in https://api.safecast.org/en-US/bgeigie_imports/awaiting_response

I can take a look today but thought I'd mention it incase the reason is obvious to you.

matschaffer:
Tried https://api.safecast.org/en-US/bgeigie_imports/33560 as well

Seems like I'm missing a message display here:

bgeigie_import__33560____bgeigie_imports____safecast_api

Could also be a side-effect of commenting on my own drives.

matschaffer:
```
safecast=# SELECT * FROM measurement_imports WHERE id IN (32161, 33560);
-[ RECORD 1 ]------+--------------------------------------------------------
id                 | 32161
user_id            | 574
source             | 23131015.LOG
md5sum             | 516cea2b6ddd3dbdcf2636b4676a5ced
type               | BgeigieImport
status             | awaiting_response
measurements_count | 7
map_id             |
status_details     | --- !ruby/hash:ActiveSupport::HashWithIndifferentAccess
                   | process_file: true
                   | import_bgeigie_logs: true
                   | compute_latlng: true
                   |
approved           | f
created_at         | 2017-10-24 05:52:07.72041
updated_at         | 2017-10-24 05:52:13.370354
name               |
description        |
lines_count        | 7
credits            |
height             |
orientation        |
cities             |
subtype            | None
comment            |
rejected           | f
rejected_by        |
-[ RECORD 2 ]------+--------------------------------------------------------
id                 | 33560
user_id            | 574
source             | 23130109.LOG
md5sum             | b1633cde451f038c76ea35c8e60264ee
type               | BgeigieImport
status             | awaiting_response
measurements_count | 78
map_id             |
status_details     | --- !ruby/hash:ActiveSupport::HashWithIndifferentAccess
                   | process_file: true
                   | import_bgeigie_logs: true
                   | compute_latlng: true
                   |
approved           | f
created_at         | 2018-01-10 08:40:14.667593
updated_at         | 2018-07-01 23:17:24.579537
name               | Random test
description        |
lines_count        | 78
credits            | Mat Schaffer
height             | 1.00
orientation        | Facing Left
cities             | Japan
subtype            | Drive
comment            |
rejected           | f
rejected_by        |
```

matschaffer:
here's the schema incase prod doesn't quite match dev

```
safecast=# \d measurement_imports
 created_at         | timestamp without time zone | not null
 updated_at         | timestamp without time zone | not null
 subtype            | measurement_imports_subtype | not null default 'None'::measurement_imports_subtype
 rejected           | boolean                     | default false
 rejected_by        | character varying(255)      |
Indexes:
```

Issue #407, opened 2017-12-03
Reject w/o metadata


seanbonner:
Currently in order to reject a drive it requires metadata to have been added, however one reason for rejection is lack of metadata and often these logs are missing all the info, so a moderator has to go and fill out all the info boxes with "none" or something then save it before they can reject it. Should be able to reject immediately.

matschaffer:
Looks like the conditional on https://github.com/Safecast/safecastapi/blob/master/app/views/bgeigie_imports/_head.html.erb#L8 should be tweakable to allow this.

Probably makes sense to move it into a `BgeigieImport#rejectable?` method as well since it's getting pretty long.

eitoball:
It should be able to reject imports without metadata. If you need to fill metadata to make reject button on import detail page, #460 should display reject button on import detail page even if import has no metadata.

matschaffer:
@eitoball so the action works, but I think maybe something in the UI is hiding it. For example here's a upload I did but never added metadata for. I'm logged in as an admin but there's no reject as far as I can see.

https://api.safecast.org/en-US/bgeigie_imports/22730

![screen_shot_2018-06-11_at_8_44_47](https://user-images.githubusercontent.com/690/41207632-06f9c1a6-6d54-11e8-9fb7-8cd33ffdc890.png)

And it shows up in the "not approved" queue.

https://api.safecast.org/en-US/bgeigie_imports/not_approved?page=8

![screen_shot_2018-06-11_at_8_45_35](https://user-images.githubusercontent.com/690/41207636-15bd27c8-6d54-11e8-934f-25a637be9100.png)

eitoball:
>  but I think maybe something in the UI is hiding it.

@matschaffer Yes, it is. #460 needs to be merged to show "Reject" button. I thought that I mentioned about that in description of #462.

matschaffer:
Ah, thanks. Guess I missed that detail. Merged #460 now and will get that pushed out.

PR #462, opened 2018-06-05
Add spec for bgeigie_imports#reject

eitoball:
## ~~This PR have to be merged after #460 is merged.~~

Connected to #407.

b611473cf173c0658cdd185485686df962e7d390 is the only commit for this PR.

It adds a spec for bgeigie_import#reject to make sure bGeigie import without metadata can be rejected.


matschaffer:
@eitoball should this get rebased? https://github.com/Safecast/safecastapi/pull/460 is merged but I still see a bunch of commits from that PR also in this list

eitoball:
> should this get rebased? #460 is merged but I still see a bunch of commits from that PR also in this list

@matschaffer Rebased. It could be able to merge, and those duplicate commits were gone as long as commit hashes are same, I believe. Anyway, thank you.

Issue #245, opened 2016-04-17
Add "Highest Measurement" field to log approval page


seanbonner:
Right now we have:

Filename
Number Of Lines
Number Of Measurements 

I'd like to add to that 

Highest Measurement

Reason being, with the map only it's too easy to miss a few very high readings on a log where 99.9% of the readings are within normal scope which leads to accidental approval regularly. If we can show "Highest Measurement" then the moderator can quickly see if there is reason to look closer at the log before approving or rejecting.

robouden:
Maybe easiest way is to use the code we have for Databot (Python).

rob



Regards,
Rob Oudendijk Yuka Hayashi
http://yr-design.biz   http://oudendijk.biz  http://about.me/robouden
tel +81 80-22605966  Skype: robouden  Facebook:robouden
<http://on.fb.me/QeKw2P> twitter:robouden <http://bit.ly/RAiSTC>

nokton:
There is supposed to be another github request to embed the Databot diagnostics into the drive import and show these on the tab Process Log - that includes peak and error counts, etc. 

Pieter

matschaffer:
Bumping to next given @nokton's [request in slack today](https://safecast.slack.com/archives/C026F2Q1U/p1527643203000156)

matschaffer:
@eitoball's fix is looking good on prod:

![bgeigie_import__10757____bgeigie_imports____safecast_api](https://user-images.githubusercontent.com/690/42573177-49d485b4-8556-11e8-98d0-39d0f7b163d2.png)

seanbonner:
Looks like there's some bugs with this. Check this drive:

https://api.safecast.org/en-US/bgeigie_imports/37388

Says highest reading is 141, but mousing over the red spots on the map shows readings over 300. Just saw several like this...

eitoball:
@seanbonner 

I am sorry for taking so long to respond.

Anyway,

> Says highest reading is 141,

The highest "logged" CPM value is `141` (`177`, actually). 

> but mousing over the red spots on the map shows readings over 300. 

Yes, there are. For example,

<img width="852" alt="Screen Shot 2022-09-27 at 9 49 39" src="https://user-images.githubusercontent.com/34205/192406207-d7fbced4-d3ca-4711-a370-62dcecfaa3c0.png">

[...]

PR #463, opened 2018-06-07
Add "Highest measurement" on bGeigie import detail page

eitoball:
Connected to #245.

bgeigie_import__1____bgeigie_imports____safecast_api

### DIFF
--- a/app/views/bgeigie_imports/_static_metadata.html.erb
+++ b/app/views/bgeigie_imports/_static_metadata.html.erb
@@ -1,28 +1,14 @@
 <div class="span2">
   <dl>
-    <dt>
-      <%= t('.uploaded_by') %>
-    </dt>
-    <dd>
-      <%= @bgeigie_import.user.name_or_email %>
-    </dd>
-    <dt>
-      <%= t('.filename') %>
-    </dt>
-    <dd>
-      <%= @bgeigie_import.source.filename %>
-    </dd>
-    <dt>
-      <%= t('.number_of_lines') %>
-    </dt>
-    <dd>
-      <%= @bgeigie_import.lines_count %>
-    </dd>
-    <dt>
-      <%= t('.number_of_measurements') %>
-    </dt>
-    <dd>
-      <%= @bgeigie_import.processed? ? @bgeigie_import.bgeigie_logs.size : '-' %>
-    </dd>
+    <dt><%= t('.uploaded_by') %></dt>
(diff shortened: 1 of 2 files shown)


app/views/bgeigie_imports/_static_metadata.html.erb
```diff
@@ -1,28 +1,14 @@
...
-    </dd>
+    <dt><%= t('.uploaded_by') %></dt>
+    <dd><%= @bgeigie_import.user.name_or_email %></dd>
+    <dt><%= t('.filename') %></dt>
+    <dd><%= @bgeigie_import.source.filename %></dd>
+    <dt><%= t('.number_of_lines') %></dt>
+    <dd><%= @bgeigie_import.lines_count %></dd>
+    <dt><%= t('.number_of_measurements') %></dt>
+    <dd><%= @bgeigie_import.processed? ? @bgeigie_import.bgeigie_logs.size : '-' %></dd>
+    <dt>Highest Measurement</dt>
```
matschaffer: Seems like we might not want to break the translation pattern here. Mind at least dropping it into the `en` strings file?

app/views/bgeigie_imports/_static_metadata.html.erb
```diff
@@ -1,28 +1,14 @@
...
+    <dd><%= @bgeigie_import.maximum_cpm || '-' %></dd>
```
matschaffer: Looks like the show page isn't bad anymore, but could be good to check on [newrelic](https://rpm.newrelic.com/accounts/1057812/applications/10580222/transactions?type=app#id=5b22436f6e74726f6c6c65722f626765696769655f696d706f7274732f73686f77222c22225d) once this is deployed to ensure the new `max()` call doesn't cause undue load.

app/views/bgeigie_imports/_static_metadata.html.erb
eitoball: Yep, done.

app/views/bgeigie_imports/_static_metadata.html.erb
matschaffer: Newrelic is looking clean. And also looks like that page doesn't see massive load anyway, so I'd say we're 👍 

screen shot 2018-07-12 at 8 22 47

eitoball: Thank you for checking out.

PR #488, opened 2018-08-13
Windows support

AkuaBiaa:


matschaffer:
Looks good @AkuaBiaa - I can bundle and run tests on my mac still so seems we've successfully added windows support w/o breaking mac support 👍

Issue #487, opened 2018-08-07
Use 0.1 step for sensor height +/- buttons


matschaffer:
From @jamoross via https://safecast.slack.com/archives/C026F2Q1U/p1533643611000117

It steps by whole numbers right now which is too big for the usual 0.5-2 range. 0.1 per click could be better.

Can probably just adjust step=0.1 here, though a cross-browser check would be good.

![bgeigie_import__20565____bgeigie_imports____safecast_api](https://user-images.githubusercontent.com/690/43780867-8f0a6a18-9a96-11e8-9e94-27ab6736f5e1.png)

PR #492, opened 2018-08-14
Sensor height update

AkuaBiaa:
Updated sensor height step from 1m to 0.1m. Hope I updated the right file...

Connected to #487

### DIFF
--- a/app/views/bgeigie_imports/_edit.html.erb
+++ b/app/views/bgeigie_imports/_edit.html.erb
@@ -70,7 +70,8 @@
 
       <%= f.input :height,
                   :input_html => {
-                    :class => 'span2'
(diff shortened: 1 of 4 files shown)


eitoball:
@AkuaBiaa I think that [`app/views/bgeigie_imports/_edit.html.erb`](https://github.com/Safecast/safecastapi/blob/master/app/views/bgeigie_imports/_edit.html.erb) is view file for creating/editing bGeigie import. It uses [simple_form gem](https://github.com/plataformatec/simple_form). This generates `step` attributes because `height` column is numeric.

AkuaBiaa:
@eitoball  Thank you! I would read more about that and hopefully be able to correct the step properly.

PR #500, opened 2018-09-17
Add Dockerfile for application

eitoball:
This PR adds `Dockerfile` for application.


eitoball:
> You using this for development?

Yep. In RubyMine, ruby in docker container can be specified as remote interpreter. It is usable, but I sometimes need to restart container.

Issue #131, opened 2013-06-08
Improve sorting of imports listings for moderation


nokton:
For https://api.safecast.org/en-US/bgeigie_imports - would be good to have the most recent uploads at the top of the list. Right now we appear to be just getting default row ordering.

Additionally we'd like to have sorting on at least the following columns:

- highest value
- has comment
- uploader _name_ (currently sorting is user id, but name would be more useful)
- number of total drives user has submitted (useful for identifying newcomers to the system)

As well as any others that are easily sortable.

Some of the above columns could be tricky as sort fields, if so we'll split out just those column sort options to their own github issues once someone has done an initial pass.

matschaffer:
Since the deploy of #501 it seems date sorting isn't happening, so put this back in to "in progress" and will try to figure out what's up

matschaffer:
@rdeskins this might work for sorting by user name

https://stackoverflow.com/a/8351908/69002

PR #501, opened 2018-09-24
Sort bgeigie imports by creation date/time

eitoball:
Connected to #131 

This PR should order bgeigie imports by creation date/time.

### DIFF
--- a/app/models/bgeigie_import.rb
+++ b/app/models/bgeigie_import.rb
@@ -15,7 +15,7 @@ class BgeigieImport < MeasurementImport # rubocop:disable Metrics/ClassLength
   has_many :bgeigie_logs, dependent: :delete_all
   has_many :uploader_contact_histories, dependent: :delete_all
 
-  scope :newest, -> { order('created_at DESC') }
+  scope :newest, ->(*) { order(created_at: :desc) }
   scope :oldest, -> { order('created_at') }
 
   %i(submitted processed done).each do |status|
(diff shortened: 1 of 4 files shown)


app/models/bgeigie_import.rb
```diff
@@ -15,7 +15,7 @@ class BgeigieImport < MeasurementImport # rubocop:disable Metrics/ClassLength
   has_many :bgeigie_logs, dependent: :delete_all
   has_many :uploader_contact_histories, dependent: :delete_all
 
-  scope :newest, -> { order('created_at DESC') }
+  scope :newest, ->(*) { order(created_at: :desc) }
```
matschaffer: First time I've seen `(*)` - why is is required just on this scope?

spec/controllers/bgeigie_imports_controller_spec.rb
```diff
@@ -19,6 +19,10 @@
 
       it 'should get all imports' do
         expect(bgeigie_imports.size).to eq(4)
+
+        # should ordered by newly created import first
+        created_at = bgeigie_imports.map { |i| i.created_at }
```
matschaffer: Looks like rubocop is calling this line out btw. Should be `map(&:created_at)` - also turned rubocop on for semaphore too ;)

eitoball: Fixed. Thx.

app/models/bgeigie_import.rb
eitoball: Because I wanted to use `:newest` as default scope, and I followed [this documentation](
https://github.com/plataformatec/has_scope#apply-scope-on-every-request)

matschaffer:
I merged & deployed this to staging but not seeing much difference on https://api-staging.safecast.org/en-US/bgeigie_imports so moving the parent ticket back to in-progress.

Issue #481, opened 2018-07-24
Shortcut for opening a full list of drives in individual tabs


matschaffer:
As part of the moderation flow Pieter works from https://api.safecast.org/en-US/bgeigie_imports and does:

1) sort - by date, newest first (as old ones are typically problem cases waiting for resolution) - by user, if someone has asked for a quick approval
2) open all drives in one page view in seprate  tabs (e.g. 25 drives)
3) approve tab by tab
4) If drive has a problem, skip for another moderator

That step 2 of getting tabs open is time consuming.

In lieu of other UX ideas, having some sort of shortcut to "open all" would save his time.

Would be good to only show this to moderators as having too many people (or a site crawler) hit this could be nasty for site load.

matschaffer:
@Lugulbanda some good thoughts here but let's keep this ticket focused on at least speeding up the current _moderation_ flow.

If you want to open up another issue with a specific action item (e.g., pre-select drive type base on users last upload) definitely feel free to do that. Don't worry if it seems like a duplicate GitHub issue. I can help dedupe since I keep regular watch over the issues.

matschaffer:
@nokton this should be all set on api.safeast.org now just make sure to watch for popup blockers

<img width="994" alt="screenshot_2018_12_13_10_48" src="https://user-images.githubusercontent.com/690/49910067-b647bd00-fec4-11e8-8eac-3f48510106ef.png">

<img width="332" alt="screen shot 2018-12-13 at 10 45 10" src="https://user-images.githubusercontent.com/690/49910073-bba50780-fec4-11e8-9b87-e2518eecb230.png">

PR #504, opened 2018-10-24
Open all button

Pheipp:
connected to https://github.com/Safecast/safecastapi/issues/481

Button that's only visible for moderators. Starts a coffeescript, that will open all unapproved drives in a new tab.


app/assets/javascripts/application/open_all.js.coffee
```diff
@@ -0,0 +1,7 @@
+jQuery ->
+
+$("#open_all").click ->
+
+  imports = $(this).data('bgeigie');
+  for dataset in imports
+    window.open("bgeigie_imports/"+dataset.id) unless dataset.approved
```
matschaffer: New line would be good here (note the 🚫 shown on github)

app/assets/javascripts/application/open_all.js.coffee
```diff
@@ -0,0 +1,7 @@
+jQuery ->
+
+$("#open_all").click ->
+
+  imports = $(this).data('bgeigie');
```
matschaffer: I'd call this "bgeigie_imports". Looks like `imports` might be some kind of reserved word based on how it's highlighting

app/views/bgeigie_imports/index.html.erb
```diff
@@ -51,6 +51,17 @@
 <%- if @bgeigie_imports.empty? -%>
   <%= t("none_found") -%>
 <%- else -%>
+
+  <ul class="inline pull-right">
+    <%-if user_signed_in? %>
+      <%- if current_user.moderator? -%>
+        <li id="open_all" data-bgeigie="<%=@bgeigie_imports.to_json%>">
```
matschaffer: Seems like it'd be better to have this be an id list, or possibly better yet get the list of IDs out of the table. My guess is this leads to a very very large json blob in the middle of the HTML.

app/views/bgeigie_imports/index.html.erb
```diff
@@ -51,6 +51,17 @@
...
+          <%=link_to t(".Open all"),params, :class => 'btn' %>
```
matschaffer: Can probably have this be a link to `"#"` instead of `params` and just avoid the navigation altogether if JS isn't enabled.

app/views/bgeigie_imports/index.html.erb
Pheipp: But if I get the list of IDs out of the table, the button has to be under the tabe right? It would be smoother for the moderation flow if the button would be over the table.

matschaffer: No you can have the ids in the erb (might be there already) then have the coffescript/javascript read them out. The javascript won't execute until clicked so page location won't matter

eitoball: Since we have `@bgeigie_imports` here, we can get list of IDs like `@bgeigie_imports.map(&:id)` (or `@bgeigie_imports.map { |i| i.id }`.  If we want to filter with unapproved imports, then `@bgeigie_imports.each_with_object([]) { |import, ary| ary << import.id unless import.approved? }`. I think that filtering should be done in Rails application side because application knows what is approved import.

Currently, URLs for import pages are generated in `open_all.js.coffee`, but it would be better to generate URLs in views or controllers and pass these URLs to JavaScript. So, it would be like:

```erb
<%= link_to t('.Open all'), '#', class: 'btn', id: 'open_all', data: { urls: @bgeigie_imports.each_with_object([]) { |import, ary| ary << bgeigie_imports_path(i) unless import.approved? }.to_json } %>
```

And, in JavaScript, you just iterate list of URLs that are passed via `data-urls` attribute , and open each.

app/views/bgeigie_imports/index.html.erb
```diff
@@ -51,6 +51,17 @@
 <%- if @bgeigie_imports.empty? -%>
   <%= t("none_found") -%>
 <%- else -%>
+
+  <ul class="inline pull-right">
+    <%-if user_signed_in? %>
+      <%- if current_user.moderator? -%>
```
eitoball: You might be able to simplify these two `if` statements with just one `if` statement like `if current_user.try!(:moderator?)`. 

`current_user` would be `nil` if user is not logged-in. So, `current_user.try!(:moderator?)` would be `nil` which is falsy value. If use is logged-in, then `current_user` would be a `User` object, and `current_user.try!(:moderator?)` would return value depending on user's status.

app/assets/javascripts/application/open_all.js.coffee
eitoball: Most of editors have options or plugins to insert new line at the end of file.

app/assets/javascripts/application/open_all.js.coffee
eitoball: Maybe, is it because of variable declaration? On line 6, `imports` is not highlighted.

Anyway, `import` is indeed reserved word in ES6 (ES2015) or later, so it might be better to avoid using `import` or `imports`.

app/views/bgeigie_imports/index.html.erb
matschaffer: Let's please use a view helper though and not pack that much Ruby into the erb 😉

app/views/bgeigie_imports/index.html.erb
matschaffer: Another place a view helper would be nice `is_moderator?` could be a good name.

app/views/bgeigie_imports/index.html.erb
matschaffer: But I agree with eito's point. The table doesn't know about moderation status so assembling a list in erb seems like probably a good move. We may want to relabel the button though. "Open all unmoderated" or something.

Pheipp:
Pushed files with suggested changes :)

app/helpers/bgeigie_imports_helper.rb
```diff
@@ -33,4 +33,16 @@ def operation_button(bgeigie_import, action, text = t(format('.%s', action)))
       f.submit text, class: 'btn btn-primary'
     end
   end
+
+  def unmoderated_id_list
```
eitoball:
```suggestion
  def unmoderated_id_list(bgeigie_imports)
    bgegie_imports.each_with_object([]) do |import, list|
      list << import.id unless import.approved?
    end
  end
```

It would be nicer not to depend on instace variable `@bgeigie_import`. Instead `unmoderated_id_list` helper should receive list of bGeigie imports like this.

We could use `Array#each_with_object` here.

app/helpers/bgeigie_imports_helper.rb
```diff
@@ -33,4 +33,16 @@ def operation_button(bgeigie_import, action, text = t(format('.%s', action)))
...
+    unmoderated_id_list = []
+    @bgeigie_imports.each do |bgeigie_import|
+      unmoderated_id_list << bgeigie_import.id unless bgeigie_import.approved?
+    end
+    unmoderated_id_list
+  end
+
+  def moderator?
```
eitoball:
```suggestion
  def moderator?(user)
```

This helper also should receive `User` object as argument.

app/views/bgeigie_imports/index.html.erb
```diff
@@ -51,6 +51,15 @@
 <%- if @bgeigie_imports.empty? -%>
   <%= t("none_found") -%>
 <%- else -%>
+
+  <ul class="inline pull-right">
+    <%- if moderator? -%>
+      <li id="open_all" data-bgeigie="<%=unmoderated_id_list%>">
```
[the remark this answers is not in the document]
Pheipp: But wasn't the list so we don't have so much json in the erb? Or is it ok this time because we only have the urls and not all the data?

app/helpers/bgeigie_imports_helper.rb
```diff
@@ -33,4 +33,14 @@ def operation_button(bgeigie_import, action, text = t(format('.%s', action)))
       f.submit text, class: 'btn btn-primary'
     end
   end
+
+  def unmoderated_id_list(bgeigie_imports)
+    bgeigie_imports.each_with_object([]) do |import, import_url_list|
+      import_url_list << bgeigie_import_path(import) unless import.approved?
```
matschaffer: Just occurred to me here that this might catch rejected as well. Will push to staging & see, but we might want to say if `import.submitted?`

Pheipp:
Yes, in my tests I opend all 25 windows that where on this site.

Pheipp:
Implemented a message if the list is all approved. Also refused files won't be opend anymore. In my tests all 25 Files n the list could be opend with the button. But apparently I broke something and can't get through the checks.

app/assets/javascripts/application/open_all.js.coffee
```diff
@@ -0,0 +1,9 @@
+
+jQuery ->
+$("#open_all").click ->
+  import_url_list = $(this).data('bgeigie');
+  if import_url_list != -1
```
eitoball: I am not sure if this is JavaScript-way or not. Why not checking if `import_url_list` is empty like `import_url_list.length == 0` (and keep `unmoderated_id_list` helper as before)?

Pheipp: well, sometimes I overthink a little bit and it gets more complicated than it needed to be ;)

matschaffer:
Whoops. Meant to click merge on another PR. But at least this is the sort of thing we can patch up later easily

Issue #470, opened 2018-06-27
Targeted, daily CSV export of bgeigie data


matschaffer:
Jeffrey Imbrogno's work has been focused on data sets for particular cities:  Uppsala, Hong Kong, Seoul, Taipei, Amsterdam, Perth, Sydney, Honolulu, and Munich

Depending on the size of the data set, sometimes the API will time out (~60k points).

I think the simplest short-term workaround will be so produce daily CSV exports for each of these cities. A 3km circle from city center should be sufficient.

Some example queries:
- Hong Kong - https://api.safecast.org/measurements.csv?distance=3000&latitude=22.316905+&longitude=114.157841
- Uppsala - https://api.safecast.org/measurements.csv?distance=3000&latitude=59.8586+&longitude=17.6389

Lugulbanda:
why is there such a focus on citys like munich ?

kbfastcat:
underground zombies...

PR #506, opened 2018-11-01
Imbrogno exports

Pheipp:
connected to https://github.com/Safecast/safecastapi/issues/470 

Imbrogno_export exports all datasets and Imbrogno_daily_export exports all datasets uploaded on that day. The radius to the citycenter is 3000 meters.


script/imbrogno_export.sql
```diff
@@ -0,0 +1,7 @@
+-- imbrogno_export and the SQL-Query imbrogno_export.sql:
+-- Exports all datasets into a csv file, where the location is within a radius of 3000 meter from
+-- the citycenter of: Uppsala, Hong Kong, Seoul, Taipei, Amsterdam, Perth, Sydney, Honolulu and Munich
+-- to prevent a timeout of the API
+SET search_path TO "$user", public, topology, postgis;
```
matschaffer: Is there a reason we're setting search_path here but not in the daily file?

Pheipp: It was just there so I don't need it to write this every time. Thought that this statement is not needed in the final version, so I put the statements out of the code. But unfortunately I forgot this one.

Issue #486, opened 2018-08-07
Provide an import ID when attempting to upload a drive with md5 collision


matschaffer:
In coding out https://gist.github.com/matschaffer/2b073e3556fc47d9c30fe4c40368ac4e for https://github.com/Safecast/safecastapi/issues/453 I noticed that if we upload the same file twice, there's no direct way to get the ID of the conflicting file.

The error is just

```json
{"errors": {"md5sum": ["has already been taken"]}}
```

The UI will provide it via an in-view lookup on https://github.com/Safecast/safecastapi/blob/c42345c7f098d2cedc9dabcd1574781f53a93e8a/app/views/bgeigie_imports/new.html.erb#L11

We should customize the md5sum collision error message in the API to also include the conflicting drive ID, possibly like:

```json
{"errors": {"md5sum": ["has already been taken by 20565"]}}
```

Or possibly something more easily parsable than that.

This might also allow us to eliminate the ad-hoc query happening in the view today.

PR #507, opened 2018-11-09
Provide import ID after a Md5sum collision 

Pheipp:
connected to https://github.com/Safecast/safecastapi/issues/486

The best way I came up with, is to change the message afterwards.

### DIFF
--- a/app/models/measurement_import.rb
+++ b/app/models/measurement_import.rb
@@ -26,6 +26,6 @@ class MeasurementImport < ActiveRecord::Base
 
   def uniqueness_of_md5sum
     bgeigie_import = self.class.unscoped.find_by(md5sum: md5sum)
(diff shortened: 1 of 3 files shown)


Pheipp:
Well, sure. I will look what I can do.

Pheipp:
Well, seems like I broke something and I have no idea what. Also that's the first time I wrote a spec so I am open for improvements.

eitoball:
There was an issue when updating instances of `BgeigieImport`. I fixed the issue and moved spec as spec of `MeasurementImport`. Also,  I squashed commits and rebased on current master branch.

config/database.yml
```diff
@@ -21,7 +21,9 @@ development:
   pool: 5
   username: safecast
   su_username: safecast
-  password: 
+  password:
+  host: postgresql
```
matschaffer: We might want to be careful about adding this. Pretty sure cloud66 uses the checked in database.yml. Should probably back out the change until we verify or move the app to AWS

PR #511, opened 2018-11-14
Update ruby to 2.3.8

eitoball:
SSIA

AWS Beanstalk support ruby 2.3.8 configuration. (https://docs.aws.amazon.com/elasticbeanstalk/latest/platforms/platforms-supported.html#platforms-supported.ruby)


matschaffer:
Ah. Looks like semaphore is 2.2.9 by default might be suppressing `.ruby-version`

```
Semaphore (58 min left): ~/safecastapi $ cat .ruby-version
cat: .ruby-version: No such file or directory
Semaphore (58 min left): ~/safecastapi $ cat > .ruby-version
2.3.8
Semaphore (57 min left): ~/safecastapi $ ruby --version
rbenv: version `2.3.8' is not installed (set by /home/runner/safecastapi/.ruby-version)
```

Looks like 2.3.8 isn't available there either.

https://semaphoreci.com/docs/languages/ruby/ruby-support-on-semaphore.html

Could be time to try semaphore 2.0 there too

matschaffer:
This also gets me wondering if it's time to just run the build inside a container:

```
Semaphore (57 min left): ~/safecastapi $ docker run --rm -it ubuntu bash
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu
32802c0cfa4d: Pull complete
da1315cffa03: Pull complete
fa83472a3562: Pull complete
f85999a86bef: Pull complete
Digest: sha256:6d0e0c26489e33f5a6f0020edface2727db9489744ecc9b4f50c7fa671f23c49
Status: Downloaded newer image for ubuntu:latest
```

.rubocop.yml
```diff
@@ -120,6 +120,87 @@ Layout/SpaceInsideParens:
...
 
+Style/FrozenStringLiteralComment:
+  Exclude:
+    - 'Gemfile'
+    - 'Rakefile'
+    - 'bin/bundle'
+    - 'bin/rails'
+    - 'bin/rake'
+    - 'config.ru'
+    - 'db/migrate/20111123174941_devise_create_users.rb'
```
matschaffer: Should we maybe disable this rule? Having to add a line here every time we add migration seems like it'd get tiring.

eitoball: For new file, magic comment should be added. I just don't want to modify old migration files.

eitoball:
> Could be time to try semaphore 2.0 there too

@matschaffer Yes, I think so. I would like to migrate to 2.0. Could you DM me access token for semaphore 2.0?

> This also gets me wondering if it's time to just run the build inside a container:

I would like to migrate to semaphore 2.0 and come back to this discussion later.

matschaffer:
Looks good, pulling this in so I can base beanstalk work on it.

PR #520, opened 2018-12-01
Refactor to use elasticsearch-model

eitoball:


spec/support/elasticsearch_helpers.rb
```diff
@@ -0,0 +1,66 @@
...
+    client.indices.delete(index: 'ingest-measurements-*')
+  end
+
+  def add_index_to_ingest_measurements(client, body)
+    client.index(
+      index: 'ingest-measurements-2018-12-01',
+      type: 'measurement',
+      body: body.merge('@timestamp': '2018-12-01T12:34:56Z')
+    )
+    # Need to have some period to get result
```
matschaffer: Interesting. I get the impression there's a method we can use on client.index to force the result

eitoball: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#index-wait-for-active-shards

Maybe this?

PR #524, opened 2018-12-02
Disallow to delete approved bGeigie imports

eitoball:
Connected to #496

### DIFF
--- a/app/views/bgeigie_imports/_show.html.erb
+++ b/app/views/bgeigie_imports/_show.html.erb
@@ -71,4 +71,4 @@
   </div>
 </div>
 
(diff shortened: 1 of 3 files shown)


app/views/bgeigie_imports/_show.html.erb
```diff
@@ -71,4 +71,4 @@
   </div>
 </div>
 
-<%= render "delete" if current_user == @bgeigie_import.user || current_user.try(:moderator?) %>
+<%= render "delete" if !@bgeigie_import.approved? && (current_user == @bgeigie_import.user || current_user.try(:moderator?)) %>
```
matschaffer: Every time I see these long authorization conditionals I wonder if we should set up cancan or something

PR #525, opened 2018-12-02
Ingest-data-View Hackathon 2018

Pheipp:
What kind of filters do we need besides area?
Date and the field(for example: PM 2.5)? maybe some other?

### DIFF
--- a/app/controllers/ingest_controller.rb
+++ b/app/controllers/ingest_controller.rb
@@ -1,15 +1,10 @@
 class IngestController < ApplicationController
-  has_scope :area
-  has_scope :field
-  has_scope :uploaded_after
-  has_scope :uploaded_before
-
   def index
+    @field = params[:field]
+    @uploaded_after = params[:uploaded_after]
+    @uploaded_before = params[:uploaded_before]
     @area = params[:area]
-    @data = []
-    IngestMeasurement.data_for(device_urn: device_ids).each do |data|
-      @data << { 'when_captured' => data[:when_captured], 'pms_pm02_5' => data[:pms_pm02_5], 'device' => data[:device] } if data['pms_pm02_5']
-    end
+    ingest_data
     respond_to do |format|
       format.html
       format.csv { generate_csv }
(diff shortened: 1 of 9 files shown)


Pheipp:
apihackathon_2

At the moment you can only search by an area. The CSV - export should work

Pheipp:
Added some fields for: 'opc_pm01_0','opc_pm02_5', 'opc_pm10_0', 'pms_pm01_0', 'pms_pm02_5', 'pms_pm10_0'

Pheipp:
For the time, should we filter it with a query or should I make a method in the controller?

eitoball:
> For the time, should we filter it with a query or should I make a method in the controller?

What do you mean by `a method in the controller`? Could you show us some pseudo code?

Pheipp:
In your query I get every measure in that area right? So I thougt to display them, I could just make an if-statement like:

```
IngestMeasurement.data_for(device_urn: device_ids).each do |data|
      @data << { :when_captured => data[:when_captured], :device => data[:device] } if data[:pms_pm02_5] and :uploaded_after >= @uploaded_after and :uploaded_before <= @uploaded_before 
    end
```
something like this

eitoball:
@Pheipp For example, like this

```ruby
IngestMeasurement.data_for(device_urn: device_ids).select do |data|
      data[:pms_pm02_5] and :uploaded_after >= @uploaded_after and :uploaded_before <= @uploaded_before 
  end.map do |data|
    data.select(:when_captured, :device)
  end
```

Pheipp:
`data.select(:when_captured, :device)` gives me an ArgumentError. Also for every field I need to extract another index. For  example: 'pms_pm01_0' for 'PM 1.0' and 'pms_pm02_5' for 'PM 2.5'. For now, I created a new hash with a value key. 
`@data << { when_captured: data[:when_captured], value: data[@field], device: data[:device] }`

Is there a way to do this with the select method?

Pheipp:
Needed to do some changes so it got pass through the checks

matschaffer:
This view is technically on prod but not yet configured with credentials to access the data. Will keep it in the "done" column until that's sorted and the view is functional.

PR #539, opened 2019-04-07
Add latest map snapshot manual SQL export

Frangible:


matschaffer:
Such a shame the hoops you have to run through here. Will make sure you have a way to run these more directly once we're on RDS.

PR #546, opened 2019-05-06
Some ops docs regarding DB overload

matschaffer:


robouden:
Seems good to me too.

robouden:
Ant reason why CircleCI failed?

matschaffer:
I'm still having some trouble getting circle to like our postgis setup

matschaffer:
Not yet. The jobs will require some work before we can make use of a replica. The iOS export will probably need to stay on the primary since it makes some data modifications and uses temp tables which unfortunately postgres won't allow on replicas

PR #556, opened 2019-06-09
Use PostgreSQL count estimate trick on dashboard page

eitoball:
Implement method that described in https://wiki.postgresql.org/wiki/Count_estimate

Since number returned by `approximate_count` is approximate, I changed to use [`number_to_human``(https://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html#method-i-number_to_human)  in view. This method will display like `1.23 Billion` for `1234567890`.

### DIFF
--- a/app/views/dashboards/show.html.erb
+++ b/app/views/dashboards/show.html.erb
@@ -40,14 +40,14 @@
       <li>
         <%= link_to [:measurements] do %>
           <em>
-            <%= number_with_delimiter current_user.measurements.count %>
+            <%= number_to_human current_user.measurements.approximate_count %>
           </em>
           <%= t('.submissions') %>
         <%- end -%>
       </li>
       <li>
         <%= link_to [:bgeigie_imports] do %>
-          <em><%= current_user.bgeigie_imports.count  %></em>
+          <em><%= number_to_human current_user.bgeigie_imports.approximate_count  %></em>
           <%= t('.bgeigie_imports') -%>
         <%- end -%>
       </li>
--- a/spec/models/measurement_spec.rb
+++ b/spec/models/measurement_spec.rb
@@ -36,4 +36,10 @@ RSpec.describe Measurement, type: :model do
       end
     end
   end
+
+  describe '#approximate_count' do
+    it 'shoulde return approximate count' do
+      expect(described_class.approximate_count).to be_a(Numeric)
+    end
+  end
 end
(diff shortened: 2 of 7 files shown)


spec/models/measurement_spec.rb
```diff
@@ -36,4 +36,10 @@
       end
     end
   end
+
+  describe '#approximate_count' do
+    it 'shoulde return approximate count' do
```
sasharevzin, line 41:
```suggestion
    it 'should return approximate count' do
```
OR just
```
 it 'returns approximate count' do
```
More info: http://www.betterspecs.org/#should

sasharevzin:
Why not just to use a cache counter (the Rails way): https://blog.appsignal.com/2018/06/19/activerecords-counter-cache.html

app/views/dashboards/show.html.erb
```diff
@@ -40,14 +40,14 @@
       <li>
         <%= link_to [:measurements] do %>
           <em>
-            <%= number_with_delimiter current_user.measurements.count %>
+            <%= number_to_human current_user.measurements.approximate_count %>
```
sasharevzin, line 43: Is `approximate_count` per-user or total?

robouden: Total is the first priority.Per-user can be added later.

sasharevzin: So we shouldn’t change it here because it will show the same count for all users

eitoball:
> Why not just to use a cache counter (the Rails way): 

@sasharevzin We are already using counter cache. ( https://github.com/Safecast/safecastapi/blob/master/app/models/measurement.rb#L12 ) 

But, as you might notice, we should display measurement counts per user, so counter cache cannot be used.

app/views/dashboards/show.html.erb
eitoball: True. We cannot use this trick...

sasharevzin:
> > Why not just to use a cache counter (the Rails way):
> 
> @sasharevzin We are already using counter cache. ( https://github.com/Safecast/safecastapi/blob/master/app/models/measurement.rb#L12 )
> 
> But, as you might notice, we should display measurement counts per user, so counter cache cannot be used.

Cool, sorry I missed it. Still learning the app)))

eitoball:
I am closing this PR because we need to count measurements per user ( https://github.com/Safecast/safecastapi/pull/556#discussion_r417979648 ). I have no idea how to count other than caching at this moment.

PR #557, opened 2019-06-11
Update rubocop to 0.71.0

eitoball:


lib/elasticbeanstalk_helper.rb
```diff
@@ -35,7 +35,7 @@ def current_environment_number
     if selected_environments.empty?
       0
     else
-      selected_environments.sort.last.split('-').last.to_i
+      selected_environments.max.last.split('-').last.to_i
```
matschaffer: Guessing this is supposed to be like this. Always sort of "felt" like `max` was number specific. But yeah, should work on strings just fine right? :)

```suggestion
      selected_environments.max.split('-').last.to_i
```

eitoball: Yes. Using `#sort` means that you compare elements. `#max` also compares each element.

PR #558, opened 2019-06-13
Fix error when bGeigie import with awaiting_response has no contact history

eitoball:


spec/views/bgeigie_imports/show.html.erb_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  context 'When BgeigieImport status is awaiting_response' do
+    before do
+      assign(:bgeigie_import, bgeigie_import)
+      params[:id] = bgeigie_import.id
+    end
+
+    context 'with no contact history' do
+      let(:bgeigie_import) { Fabricate(:bgeigie_import, status: 'awaiting_response') }
+
+      it '例外が発生しない' do
```
matschaffer: おっ 😉

matschaffer:
@eitoball HEADS UP! a rebase seemed to fix this, so I force pushed it

Issue #550, opened 2019-05-22
Increase nginx timeout


matschaffer:
Having some trouble loading pages (e.g., /admin) now that we're on beanstalk with default 60s nginx time out.

We should bump that up. ELB should be 300s already.

PR #568, opened 2019-07-23
Set ELB & nginx timeout to 600s

matschaffer:
Fixes https://github.com/Safecast/safecastapi/issues/550

### DIFF
--- a/.ebextensions/web.config
+++ b/.ebextensions/web.config
@@ -1,0 +1,12 @@
+option_settings:
+  - namespace: aws:elb:policies
+    option_name: ConnectionSettingIdleTimeout
+    value: 600
+
+commands:
+  set_web_tier_timeout:
+    command: |
+      if ! /opt/elasticbeanstalk/bin/get-config meta -k sqsdconfig 2>/dev/null; then
+        echo "proxy_read_timeout 600s;" > /etc/nginx/conf.d/web.conf
+        service nginx restart
+      fi
--- a/.ebextensions/web.config
+++ b/.ebextensions/web.config
@@ -4,7 +4,7 @@ option_settings:
     value: 600
 
 commands:
-  set_web_tier_timeout:
+  match_nginx_timeout_to_elb_timeout:
     command: |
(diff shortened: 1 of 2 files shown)


matschaffer:
Currently pushed to the dev web server https://us-west-2.console.aws.amazon.com/elasticbeanstalk/home?region=us-west-2#/environment/events?applicationName=api&environmentId=e-bmzgsz6cik

robouden:
Seems fine for me.

.ebextensions/web.config
```diff
@@ -0,0 +1,12 @@
+option_settings:
+  - namespace: aws:elb:policies
+    option_name: ConnectionSettingIdleTimeout
+    value: 600
+
+commands:
+  set_web_tier_timeout:
+    command: |
+      if ! /opt/elasticbeanstalk/bin/get-config meta -k sqsdconfig 2>/dev/null; then
```
auspicacious: I'm confused about this check. If a configuration for the Amazon simple queue service does not exist, then write a value into the Nginx configuration?

Also, not being familiar with Elastic Beanstalk, should I be concerned about idempotency? In other words, should there be a mechanism to prevent this same line from being written to the `web.conf` file twice?

matschaffer: So far it's the most concise way I've found to tell if you're in a web or worker tier. Only the workers get that sqs setting so if it's not there you're in a web tier env

auspicacious: Could you comment the code to explain that? I did searches on Google and GitHub and found very few references to that key, let alone why one would use it in that way. Thanks.

Also, is idempotency a potential issue? Could this script run twice on the same VM?

matschaffer: I use it in https://github.com/Safecast/safecastapi/blob/master/.ebextensions/workers.config#L34 too so now you've got me wondering if I should look for a better way :)

And the command is idempotent (overwrites the file in question). It runs once per deploy which is sort of what you want since you might decide to change the ELB timeout which you then want nginx to pick up on.

matschaffer: Went with some comments :)

PR #581, opened 2019-10-04
Update ruby to 2.4.9

eitoball:


.github/workflows/ruby.yml
```diff
@@ -0,0 +1,50 @@
+name: Ruby
+
+on: [push]
```
matschaffer: @eitoball you can drop this file if you want and just rely on circle for now. Github actions can't take AWS keys from forked PRs, so we'll have to re-vamp the build/deploy model to use github artifacts before this is gonna work very nicely.

db/structure.sql
```diff
@@ -2,20 +2,32 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 11.5
--- Dumped by pg_dump version 11.5
+-- Dumped from database version 9.5.14
+-- Dumped by pg_dump version 9.5.14
```
matschaffer: ping @auspicacious on the state of api psql DB version. Pretty sure it's not on 11 yet.

PR #582, opened 2019-10-04
Add a bit of help text

matschaffer:
Courtesy of @Azby


matschaffer:
@eitoball care to recommend a Japanese wording?

PR #583, opened 2019-10-05
Auto corrected by following Lint Ruby MagicComment

thinrope:
Auto corrected by following Lint Ruby MagicComment

Click [here](https://awesomecode.io//repos/Safecast/safecastapi/ruby_config_groups/1001) to configure it on awesomecode.io

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # This file is used by Rack-based servers to start the application.
 
 require ::File.expand_path('../config/environment', __FILE__)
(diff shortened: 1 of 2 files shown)


deploy/after_restart.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
eitoball: This file seems to be shell script even though `.rb` suffix.

config.ru
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
eitoball: This file is generated by rails, so we ignore to check with rubocop in `.rubocop.yml`. I would like to keep this file as is.

auspicacious:
Closing this request as these changes do not seem to be correct. Please re-open if necessary.

Issue #527, opened 2018-12-06
Drive auto-approval


matschaffer:
There are some cases where we might want to automate some approval analysis:

- drive contains no measurements where CPM=0
- drive contains no measurements where CPM>90 (around 0.3uSv)
- no excessive invalid GPS (e.g. less than 10% of drive total)
- submission is from a bGeigie ID for which we have approved 10 drives in the past
- no rejected drives for that bGeigie ID in the past year

For starters we'll want to run this query and see what matches.

Once we have some confidence we'll likely want to make this a scheduled automatic job.

This is especially desirable in the case where we can rollback approval.

One possible refinement is to exclude the area around Daiichi in a circle of 80km.

We should prefer to look at the bGeigie ID over the user ID is that we have quite a few accounts used where volunteers use multiple bGeigies, so we keep an eye on new devices before auto approving.

matschaffer:
Noting that we can't do `no excessive invalid GPS (e.g. less than 10% of drive total)` until the linked issue regarding invalid GPS is addressed.

Afaict, we discard the GPS validity info during drive file import.

nokton:
Hmmm - you’re saying that the valid flags for CPM and GPS are not stored after import to the API? I thought these are stored. 

Best,
Pieter

shoghicp:
I'm interesting on contributing. I can take a look into the stack and see what you have, will try to get a dev environment up and running.

matschaffer:
Welcome @shoghicp ! We'd certainly appreciate the help. The wiki has info on setting up local environments and we can send you a slack invite if you need any real time help.

shoghicp:
If you have Slack, that would be great. Will have time over weekends to look at this, if you want to send an invite it's (same as username) at gmail.com

matschaffer:
@shoghicp 	we had a new helper join (@ZacharyCChang0828 	) so figure he'll take point to try to move this issue forward. Of course feel free to help out where you can.

matschaffer:
Noting this test result from @ZacharyCChang0828 

```
[...]

PR #587, opened 2019-10-24
Auto approve

ZacharyCChang0828:
A new 'Auto approve" tab is added at bgeigie_import page
The name of the tab changes depends on if the drive will be auto-approved or not 
Two columns would_auto_approve and auto_apprv_no_zero_cpm are added to the bgeigie_imports table
The columns are updated upon the 'process' button is pressed
The rule 'contains no cpm=0' is currently implemented to check if the drive is auto approvable
The 

Connected to https://github.com/Safecast/safecastapi/issues/527


.gitignore
```diff
@@ -7,7 +7,9 @@ tmp/
 .sass-cache/
 *.swp
 *.swo
+*.save
 public/uploads/*
+public/Chang/*
```
eitoball: You might want to ignore your personal files in `.git/info/exclude`.

app/models/bgeigie_import.rb
```diff
@@ -364,4 +365,30 @@ def nmea_to_lat_lng(latitude_nmea, _north_south_indicator, longitude_nmea, _east
...
+
+  def minimum_cpm
+    # used to check if the drive contains cpm value lower than or equal to 0
+    bgeigie_logs.minimum(:cpm)
+  end
+
+  def check_auto_approve
+    # run each auto approval rule and
+    # update would_auto_approve column based on if all rules passed
+    ap_no_zero_cpm
```
eitoball: You might want to use `update_attributes` instead of `update_column` because (1) `update_attributes` updates `updated_at` column and (2) can update multiple columns.

```ruby
update_attributes(
  would_auto_approve: auto_apprv_no_zero_cpm,
  auto_apprv_no_zero_cpm: minimum_cpm.positive?
)
```

app/views/bgeigie_imports/_auto_approve_status.html.erb
```diff
@@ -0,0 +1,23 @@
+<ul class="unstyled nav" id="progress" >
+  <li class="nav-header">
+    <%= "Status" %>
+  </li>
+  <li>
+    <%- if @bgeigie_import.auto_apprv_no_zero_cpm -%>
```
eitoball: You might want to create view helper method to simplfy. For example,

```erb
<%= t('.no_cpm_equals_zero') %>
<%= cpu_equals_zero_icon(@bgeigie_import) %>
```

and, in view helper,

```ruby
def cpu_equals_zero_icon(import)
  content_tag(:i, class: import.auto_apprv_no_zero_cpm ? 'icon-ok pull-right' : 'icon-remove pull-right')
end
```

app/views/bgeigie_imports/_edit.html.erb
```diff
@@ -98,7 +98,8 @@
       %>
       <%= f.input :subtype,
                   :collection => %w(Drive Surface Cosmic None),
-                  :include_blank => false,
+                  :selected => "Drive",
+		  :include_blank => false,
```
eitoball: You might want to fix indentation here.

matschaffer:
```suggestion
                  :include_blank => false,
```

You know github has a button for that now @eitoball ?

db/migrate/20191017054450_add_would_auto_approve_to_measurement_imports.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddWouldAutoApproveToMeasurementImports < ActiveRecord::Migration
+  def up
+    add_column :measurement_imports, :would_auto_approve, :boolean, default: false
```
eitoball: Adding a column with a default value to an existing table might cause the entire table to be rewritten. `measurement_imports` is quite big, I think. So, you might want to do like:

```ruby
add_column :measurement_imports, :would_auto_approve, :boolean,
change_column_default :measurement_imports, :would_auto_approve, false)
```

db/migrate/20191022080113_add_cpm_equals_zero_to_measurement_imports.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddCpmEqualsZeroToMeasurementImports < ActiveRecord::Migration
+  def up
+    add_column :measurement_imports, :no_zero_cpm, :boolean
```
eitoball: I wonder if we need default value here.

app/views/bgeigie_imports/_edit.html.erb
eitoball: Ah, yes.

.gitignore
auspicacious: We should probably open a bug for this issue -- the app shouldn't be automatically creating personal directories during the import process, right?

app/models/bgeigie_import.rb
ZacharyCChang0828: I had some trouble using the update_attributes method so I use two update_attribute instead

db/migrate/20191017054450_add_would_auto_approve_to_measurement_imports.rb
ZacharyCChang0828: is it fine to edit migrate files that had already been migrated?

.gitignore
matschaffer: yeah, would be good to open an issue I apparently took @eitoball 's approach w/o even thinking about it

```
❯ cat ~/safecast/safecastapi/.git/info/exclude
# git ls-files --others --exclude-from=.git/info/exclude
# Lines that start with '#' are comments.
# For a project mostly in C, the following would be a good set of
# exclude patterns (uncomment them if you want to use them):
# *.[oa]
# *~

.envrc
public/mat/
```

db/migrate/20191017054450_add_would_auto_approve_to_measurement_imports.rb
matschaffer: @ZacharyCChang0828 yeah, especially when your local db is the only place it's been applied.

`rake db:migrate:redo` will redo your last one. Details here https://edgeguides.rubyonrails.org/active_record_migrations.html#rolling-back

ZacharyCChang0828:
I think I might need some help with the spec error test

eitoball:
> I think I might need some help with the spec error test

@ZacharyCChang0828 

It should be fixed in https://github.com/Safecast/safecastapi/pull/587/commits/992acfc0bacaea9de623f8dff4f1d5fb3b44e517

app/views/bgeigie_imports/_auto_approve_status.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      <%= t('.no_cpm_equals_zero') %>
+      <i class="icon-ok pull-right"></i>
+    <%- else -%>
+      <%= t('.no_cpm_equals_zero') %>
+      <i class="icon-remove pull-right"></i>
+    <%- end -%>
+  </li>
+  <li>
+    <%- if @bgeigie_import.would_auto_approve -%>
+      <%= t('would_auto_approve') %>
```
eitoball: Do we have this in at least `US-en.yml` ?

app/views/bgeigie_imports/_auto_approve_status.html.erb
eitoball: For example, https://github.com/Safecast/safecastapi/pull/587/commits/ebd002df4a7507f7dfae776d22eb0f71d3df20e9

eitoball:
@ZacharyCChang0828 Would it be okay if I re-order and merge commits? Or, would you?

PR #592, opened 2019-11-18
Fix db/structure.sql

eitoball:
### Before

I get error with `relation "schema_migrations" does not exist`.

```
$ docker-compose up -d
$ docker-compose exec -e 'DATABASE_URL=postgis://safecast:safecast@postgresql/safecast_development?schema_search_path=pubilc,postgis&encoding=unicode' app bundle exec rake db:drop db:create db:structure:load
...
psql:/src/db/structure.sql:1212: ERROR:  relation "schema_migrations" does not exist
LINE 1: INSERT INTO schema_migrations (version) VALUES ('20191023060...
                    ^
psql:/src/db/structure.sql:1214: ERROR:  relation "schema_migrations" does not exist
LINE 1: INSERT INTO schema_migrations (version) VALUES ('20191025034...
                    ^
GRANT
GRANT
```

### After

```
$ docker-compose up -d
$ docker-compose exec -e 'DATABASE_URL=postgis://safecast:safecast@postgresql/safecast_development?schema_search_path=public,postgis&encoding=unicode' app bundle exec rake db:drop db:create db:structure:load
...
INSERT 0 1
INSERT 0 1
GRANT
GRANT
```

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2,19 +2,31 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 11.5
--- Dumped by pg_dump version 11.5
+-- Dumped from database version 9.5.14
+-- Dumped by pg_dump version 9.5.14
 
 SET statement_timeout = 0;
 SET lock_timeout = 0;
-SET idle_in_transaction_session_timeout = 0;
 SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
 SELECT pg_catalog.set_config('search_path', '', false);
 SET check_function_bodies = false;
-SET xmloption = content;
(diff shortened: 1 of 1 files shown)


matschaffer:
I notice your description has "pubilc" instead of "public" - not sure if that would cause any issue, but mention it just incase.

matschaffer:
Yeah... I _think_ that might actually be the problem you're hitting. Since the export switches it to be fully specified as `public.schema_migrations`

eitoball:
> I notice your description has "pubilc" instead of "public" - not sure if that would cause any issue, but mention it just incase.

@matschaffer It's a typo. Fixed

PR #597, opened 2019-11-29
Auto approve

ZacharyCChang0828:
hide tab

### DIFF
--- a/app/views/bgeigie_imports/show.html.erb
+++ b/app/views/bgeigie_imports/show.html.erb
@@ -34,6 +34,7 @@
               </a>
             </li>
           <%- end -%>
+          <%- if @bgeigie_import.approved? && @bgeigie_import.would_auto_approve.present? -%>
           <li>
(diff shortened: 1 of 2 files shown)


app/views/bgeigie_imports/show.html.erb
```diff
@@ -34,15 +34,17 @@
...
-          <li>
-            <a href="#auto-approve-status" data-toggle="tab">
-              <%- if @bgeigie_import.would_auto_approve == true -%>
-                <%= t('.auto_approve_true') %>
-              <%- else -%>
-                <%= t('.auto_approve_false') %>
-              <%- end -%>
-            </a>
-          </li>
+          <%- if @bgeigie_import.approved? && !@bgeigie_import.would_auto_approve.nil? -%>
```
matschaffer:
```suggestion
          <%- if @bgeigie_import.processed? && !@bgeigie_import.would_auto_approve.nil? -%>
```

Seems like we want processed here rather than approved. Will mention it to you when you get back.

PR #607, opened 2019-12-03
Add snapshot export script

Frangible:


script/manual_exports/dump_measurements_ios_periods.sh
```diff
@@ -35,7 +35,10 @@ rm -f /tmp/ios13_32_*
 #/bin/sleep 60s
 #psql -U safecast --host $POSTGRESQL_ADDRESS safecast -f /var/deploy/api.safecast.org/web_head/current/script/manual_exports/2018-03-10_2018-09-10.sql
 #/bin/sleep 60s
-psql -U safecast --host $POSTGRESQL_ADDRESS safecast -f /var/deploy/api.safecast.org/web_head/current/script/manual_exports/2018-09-10_2019-03-10.sql
+#psql -U safecast --host $POSTGRESQL_ADDRESS safecast -f /var/deploy/api.safecast.org/web_head/current/script/manual_exports/2018-09-10_2019-03-10.sql
+#/bin/sleep 60s
+
+psql -U safecast --host $POSTGRESQL_ADDRESS safecast -f /var/deploy/api.safecast.org/web_head/current/script/manual_exports/2019-03-10_2019-09-10.sql
```
matschaffer:
```suggestion
PGOPTIONS=--search_path=public,postgis psql -f /var/app/current/script/manual_exports/2019-03-10_2019-09-10.sql
```

matschaffer: Could also export `PGOPTIONS` in https://github.com/Safecast/safecastapi/blob/master/.ebextensions/postgresql.config#L12 rather than setting it here.

Frangible: There's also an rsync line at the end, is that c66-specific or were the instance names preserved and such?

matschaffer: the
```

cp /tmp/ios13_32_*.sqlite /var/deploy/api.safecast.org/web_head/shared/system/
rm -f /tmp/ios13_32_*
rsync -Havq --bwlimit 10000 /var/deploy/api.safecast.org/web_head/shared/system/ios13_32_* reindeer.api.safecast.org.c66.me:/var/deploy/api.safecast.org/web_head/shared/system/
```

should be
```
cp /tmp/ios13_32_*.sqlite /var/app/current/public/system
rm -f /tmp/ios13_32_*
```

Sorry I didn’t put it here but GitHub only lets me suggest changes to the initial diff

matschaffer: No need for rsync since we have a shared drive now

Frangible:
Updated to remove rsync.

PR #610, opened 2019-12-06
Set LANG to en_US.UTF-8 in CircleCI container

eitoball:
To fix error in one of specs.

```
rspec ./spec/controllers/rails_admin/main_controller_spec.rb:82 # RailsAdmin::MainController login as user GET #dashboard should redirect to "/en-US"
```

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,4 +1,5 @@
 development:
+
   adapter: postgis
   schema_search_path: "public,postgis"
   script_dir: /usr/local/share/postgis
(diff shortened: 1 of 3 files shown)


config/database.yml
```diff
@@ -8,6 +9,8 @@ development:
   username: safecast
   su_username: safecast
   password:
+  host: postgresql
+  port: 5432
```
auspicacious: I haven't checked, but would this change break developers? I run Postgres on localhost.

PR #612, opened 2019-12-10
Tune timing of export jobs

matschaffer:
They've been bunching up lately (see https://grafana.safecast.cc/d/W7c552kZz/api-overview). Running the ios export every 8 hours should be okay per @Frangible 

image

Also adjusting the timing of the other jobs should help avoid them running before dump_measurements_ios is finished.

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -2,13 +2,13 @@ version: 1
 cron:
   - name: dump_clean
     url: /cron
-    schedule: "00 02 * * *"
+    schedule: "00 03 * * *"
   - name: dump_measurements
     url: /cron
-    schedule: "00 01 * * *"
+    schedule: "00 02 * * *"
   - name: dump_measurements_ios
     url: /cron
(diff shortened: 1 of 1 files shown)


cron.yaml
```diff
@@ -2,13 +2,13 @@ version: 1
...
     url: /cron
-    schedule: "00 01 * * *"
+    schedule: "00 02 * * *"
   - name: dump_measurements_ios
     url: /cron
-    schedule: "33 */4 * * *"
+    schedule: "33 */8 * * *"
   - name: imbrogno_export
     url: /cron
     schedule: "00 02 * * 7"
```
auspicacious: Will there be an unacceptable additional load on Sundays from these two jobs?

matschaffer: I don't suspect so. Last I looked the imbrogno_export was pretty light (just a few minutes) compared to the others (nearly an hour).

PR #613, opened 2019-12-10
Remove flakey test

matschaffer:
Going mildly nuclear here, but this test seems to fail roughly 50% of the time and is making deployments difficult so opening this as a possible option. Welcome to hear other options.

@eitoball & @auspicacious - let me know your thoughts


matschaffer:
Yeah, I have a partially complete implementation that asserts on `['/', '/en-US/']` but seemed better to just ditch it until we have a more reliable test.

Or even better kill the locale from the page path and use the `Accept-Language` header.

PR #618, opened 2019-12-17
Set nginx max body size to 10M

matschaffer:
We had a few >1MB uploads come in but nginx defaults to 1M max


matschaffer:
Also added a swapfile since it looks like the the nokogiri bump in https://github.com/Safecast/safecastapi/pull/616 requires more ram than our t2.smalls (2g) have available.

PR #622, opened 2019-12-27
Device Story Retry

eitoball:
References: #606 

## To fetch and update device metadata

```sh
$ bundle exec rake device_story:update_metadata
```

It seems that there is one device that does not have `device_urn`. In this PR, `device_urn` is considered to be a unique key to determine devices.

### DIFF
--- a/app/models/device_story/device.rb
+++ b/app/models/device_story/device.rb
@@ -14,7 +14,9 @@ module DeviceStory
     def update_from_metadata
       self.device_id = metadata['device']
       points = metadata.values_at('last_lon', 'last_lat')
-      self.last_location = "POINT(#{points.join(' ')})" if points.all?(&:present?)
+      raise ActiveRecord::RecordInvalid, 'Missing longitude or latitude' if points.any?(&:blank?)
+
+      self.last_location = "POINT(#{points.join(' ')})"
       self.last_values = metadata['last_values']
     end
   end
--- a/app/models/device_story/device.rb
+++ b/app/models/device_story/device.rb
@@ -6,5 +6,16 @@ module DeviceStory
 
     validates :device_urn, presence: true, uniqueness: true
     validates :metadata, presence: true
+
+    before_save :update_from_metadata
+
+    private
+
+    def update_from_metadata
+      self.device_id = metadata['device']
+      points = metadata.values_at('last_lon', 'last_lat')
+      self.last_location = "POINT(#{points.join(' ')})" if points.all?(&:present?)
+      self.last_values = metadata['last_values']
+    end
   end
 end
--- a/app/models/device_story/device.rb
+++ b/app/models/device_story/device.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module DeviceStory
+  class Device < ApplicationRecord
+    self.table_name = 'device_story_devices'
+
+    validates :device_urn, presence: true, uniqueness: true
+    validates :metadata, presence: true
+  end
+end
--- a/app/safecast/tasks/device_story/update_metadata.rb
+++ b/app/safecast/tasks/device_story/update_metadata.rb
@@ -43,19 +43,13 @@ module Tasks
 
       def parse(json)
         JSON.parse(json)
-      rescue Psych::SyntaxError
-        warn 'Could not parse device metadata.'
-        []
       end
 
       def fetch(uri)
         res = Net::HTTP.get_response(uri)
-        if res.is_a?(Net::HTTPOK)
-          res.body
-        else
-          warn 'Could not fetch device metadata from ttserve.'
-          '{}'
-        end
+        raise "Could not fetch device metadata from ttserve. #{res}" unless res.is_a?(Net::HTTPOK)
+
+        res.body
       end
 
       def devices_uri
--- a/app/safecast/tasks/device_story/update_metadata.rb
+++ b/app/safecast/tasks/device_story/update_metadata.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+module Tasks
+  module DeviceStory
+    class UpdateMetadata
+      class << self
+        def call
+          new.call
+        end
+      end
+
+      def call
+        each_metadata(parse(fetch(devices_uri))) do |device_urn, metadata|
+          begin
+            find_or_update_device(device_urn, metadata)
+          rescue ActiveRecord::RecordInvalid
+            warn "Could not update device metadata for #{device_urn}."
+          end
+        end
+      end
+
+      private
+
+      def find_or_update_device(device_urn, metadata)
+        device = ::DeviceStory::Device.find_by(device_urn: device_urn)
+        if device
+          device.update!(metadata: metadata)
+        else
+          ::DeviceStory::Device.create!(device_urn: device_urn, metadata: metadata)
+        end
+      end
+
+      def each_metadata(metadatas)
+        metadatas.each do |metadata|
+          device_urn = metadata['device_urn']
+          unless device_urn
+            warn 'Could not find device_urn.'
+            next
+          end
+          yield device_urn, metadata
+        end
+      end
+
+      def parse(json)
+        JSON.parse(json)
+      rescue Psych::SyntaxError
+        warn 'Could not parse device metadata.'
+        []
+      end
+
+      def fetch(uri)
+        res = Net::HTTP.get_response(uri)
+        if res.is_a?(Net::HTTPOK)
+          res.body
+        else
+          warn 'Could not fetch device metadata from ttserve.'
+          '{}'
+        end
+      end
+
+      def devices_uri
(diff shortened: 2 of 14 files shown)


app/safecast/tasks/device_story/update_metadata.rb
```diff
@@ -0,0 +1,66 @@
...
+            find_or_update_device(device_urn, metadata)
+          rescue ActiveRecord::RecordInvalid
+            warn "Could not update device metadata for #{device_urn}."
+          end
+        end
+      end
+
+      private
+
+      def find_or_update_device(device_urn, metadata)
```
auspicacious: I think this PR is a good idea -- start small, and build up functionality.

However, after seeing the problems with realtime, I think that this method for updating metadata might cause database performance problems that we could easily avoid up front.

If I understand it correctly, some of the metadata changes very often, and some of the metadata changes infrequently.

I suggest that when parsing the metadata, we create two separate objects, splitting out the metadata into these two groups. We should cache these objects somewhere.

Each time we fetch new data from ttserve, we compare the cached objects against the new objects and figure out exactly what we need to update in the DB. Postgres now has upsert syntax, so we can roll up all of the updates into a single SQL statement and avoid having to run a lot of UPDATEs.

https://stackoverflow.com/questions/1109061/insert-on-duplicate-update-in-postgresql

Maybe we can change ttserve output as well to make this easier?

auspicacious: On second thought, I may have been too excessive for the relatively small number of queries here.

app/models/device_story/device.rb
```diff
@@ -0,0 +1,20 @@
...
+    validates :device_urn, presence: true, uniqueness: true
+    validates :metadata, presence: true
+
+    before_save :update_from_metadata
+
+    private
+
+    def update_from_metadata
+      self.device_id = metadata['device']
+      self.last_location = "POINT(#{metadata['last_lon'] || 0} #{metadata['last_lat'] || 0})"
```
auspicacious: I'm sure that we don't have a device at `0,0`.

However, `0,0` are valid coordinates. Is it possible to instead set `last_location` to null? In general, using valid values to indicate a lack of data causes confusion in systems.

app/safecast/tasks/device_story/update_metadata.rb
```diff
@@ -0,0 +1,66 @@
...
+        device = ::DeviceStory::Device.find_by(device_urn: device_urn)
+        if device
+          device.update!(metadata: metadata)
+        else
+          ::DeviceStory::Device.create!(device_urn: device_urn, metadata: metadata)
```
auspicacious: If we move to Rails 6 we can use `upsert` which is much more reliable in multi-threaded environments :)

app/safecast/tasks/device_story/update_metadata.rb
```diff
@@ -0,0 +1,66 @@
...
+            warn 'Could not find device_urn.'
+            next
+          end
+          yield device_urn, metadata
+        end
+      end
+
+      def parse(json)
+        JSON.parse(json)
+      rescue Psych::SyntaxError
```
auspicacious: This may be a difference in programming style, but this seems to quietly hide an exception. Is there any harm in allowing the exception to propagate? I feel like we might also get better debugging information that way -- if `SyntaxError` contains any useful information, it is lost here.

I may be thinking about this in terms of Java unchecked exceptions.

app/safecast/tasks/device_story/update_metadata.rb
```diff
@@ -0,0 +1,66 @@
...
+        []
+      end
+
+      def fetch(uri)
+        res = Net::HTTP.get_response(uri)
+        if res.is_a?(Net::HTTPOK)
+          res.body
+        else
+          warn 'Could not fetch device metadata from ttserve.'
+          '{}'
```
auspicacious: Similarly, if there is an error fetching the metadata, shouldn't this raise an exception instead of silently continuing?

auspicacious: Also, ttserve usually returns a list -- `[]`. But this is an object -- `{}`. So any code trying to parse this as a list will fail.

db/migrate/20191215115843_create_device_story_devices.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class CreateDeviceStoryDevices < ActiveRecord::Migration
+  def change
+    create_table :device_story_devices do |t|
+      t.references :user
+
+      t.string :device_urn, null: false, default: ''
+      t.jsonb :metadata, null: false, default: '{}'
```
auspicacious: I understand why these are marked `NOT NULL`. However, I don't think that they should have defaults. We would never want to have a table row where the device URN is the empty string, or the metadata is completely empty. Those are error states, not valid data.

app/models/device_story/device.rb
robouden: 0,0 means that it is located in the middle of the ocean. I would consider it not valid. It indicated no GP coordinates are sent.

auspicacious: A human being might look at `0,0` and understand that it's not valid. But a computer won't know that without special programming. There are many examples of problems caused by this, the most famous of which is this:

https://splinternews.com/how-an-internet-mapping-glitch-turned-a-random-kansas-f-1793856052

In computing, we already have standard concepts for the idea of "no data" -- `null`, `undefined`, `nil`, etc. I am not aware of any reason to make things more complicated by taking a real location and making it the "no data" location.

robouden: The reason for not allowing 0,10 for me is that if a string of coordinates is being fed into a map in Grafana or Kibana, the software will show 00 and the cluster of the other coordinates. That means often as a wide scaling.

auspicacious: I think we're agreeing with each other.

eitoball: I am not sure why only `0,0` is being mentioned. In the worst case, we could have `0,0`. But, how often?

We have `metadata` column. We should look at this column if we need to look seriously. I believe that `last_*` columns are for peeking without digging matadata.

Anyway, I changed to set `last_location` only if both `last_lat` and `last_lot` are present.

db/migrate/20191215115843_create_device_story_devices.rb
eitoball: I see. I agree that we should not set default values. I removed `default` clauses. For `metadata` column, I removed `null` constraint as well. I thought that there could be a case that we would want to create device page before setting it up. In that case, we would not have metadata because device would not be working yet.

app/safecast/tasks/device_story/update_metadata.rb
eitoball: In this case, it would do nothing, just outputs a warning. `JSON.parse(`{}')` returns an empty `Hash`. `Hash` includes `Enumerable` which responds to `#each`. 

This `#fetch` method would raise exception in case of exceptional cases such as network is not available or so. ttserve is simple HTTP server. It usually returns only `200 OK`. I assume that there is an ALB in front of ttserve. If so, ALB sometimes returns `503 Bad gateway`.  In case of getting 503, I think that it would be better to continue silently.

auspicacious: Could you please explain why it would be better to continue silently? There is no data to process, only an error. Shouldn't we either retry the request or exit?

app/models/device_story/device.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    validates :device_urn, presence: true, uniqueness: true
+    validates :metadata, presence: true
+
+    before_save :update_from_metadata
+
+    private
+
+    def update_from_metadata
+      self.device_id = metadata['device']
```
auspicacious: Scenario to consider:

New device metadata arrives, but the metadata has no points. In this scenario, last_location will remain populated with the old last_location, however, the database row will still be updated.

For last_values, if there is no last_values, `self.last_values` will correctly be set to `nil`. For consistency, I think `self.last_location` should also be set to `nil` in this situation.

matschaffer: per today's recording, let's just not update if we don't have a lat/lon

@jamoross makes a good point that we probably shouldn't store 0,0 either since it's likely wrong. Or at very least we should log it or something.

app/safecast/tasks/device_story/update_metadata.rb
matschaffer: Per recording today 👍 on @auspicacious 's call for raising rather than warning and creating an empty object (on non-200) or list (on non-json).

I presume we'll be calling this on something like a 15m schedule, so it's okay to just let it explode and we'll try again in the next run.

matschaffer:
Ahhh... the device_id is a column. I think it needs `limit: 8` set to handle these ids

auspicacious:
Thanks for making those changes.

PR #626, opened 2020-01-22
Expanded all elastic beanstalk root disks to 32 GB and fixed a bug in…

auspicacious:
… the `elasticbeanstalk_helper`.

This should give us plenty of disk space. However, it does so on both the web and worker tier. I see that there's a `workers.config` but I'm not sure how that gets targeted to the workers specifically, and also if we can target dev-wrk and prd-wrk separately, since prd probably needs more disk space.

I've selectively used this code already to expand the worker nodes, but before we merge it I think we should take a look and see if we can improve it to that end.

Also worth noting that I didn't have any luck expanding the disk size by modifying the EB config files and updating existing environments, then destroying/re-creating nodes. I'm not sure if it would've worked if I created an entirely new environment; however, it's probably better to put it in `.ebextensions` anyway so it's version controlled.

### DIFF
--- a/.ebextensions/00-expand-root-disk.config
+++ b/.ebextensions/00-expand-root-disk.config
@@ -1,0 +1,10 @@
+---
+Resources:
+  AWSEBAutoScalingLaunchConfiguration:
+    Type: AWS::AutoScaling::LaunchConfiguration
+    Properties:
+      BlockDeviceMappings:
+        - DeviceName: /dev/xvda
+          Ebs:
+            VolumeSize:
+              32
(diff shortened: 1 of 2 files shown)


robouden:
Approved!!!

auspicacious:
The change seems to have worked in production -- we're now only using 20% of available disk space so there should be plenty of space for the jobs to run.

However, if I `eb config get prd-wrk` there's no changes to the Elastic Beanstalk config stored in S3. I'm not sure if it's possible to enlarge the root disk size using that config instead of a file in `.ebextensions`. I didn't have any luck when updating the config of an existing environment and re-creating instances, however, I didn't try creating an entirely new environment with a modified config.

Therefore, for expediency's sake, I'm going to merge this as-is and in the future both our web and worker tier EC2 instances will get 32 GB of root disk space. We can see if this costs a lot of money, but my estimate right now is that the bill will go up by $10 per month total.

PR #632, opened 2020-02-10
Hide unconfirmed user name on show

matschaffer:


matschaffer:
I was thinking just incase a user somehow finds their own display name before confirming this might be less confusing.

PR #634, opened 2020-02-12
Refactor docker environment

eitoball:


matschaffer:
👍👍 for this. The builds (and docker-compose’s less than great support for updating them) have been seeming like unnecessary overhead.

PR #638, opened 2020-02-21
version number which contains words now shows

ZacharyCChang0828:
Now take everything after format= as version.

### DIFF
--- a/app/models/bgeigie_import.rb
+++ b/app/models/bgeigie_import.rb
@@ -97,7 +97,7 @@ class BgeigieImport < MeasurementImport # rubocop:disable Metrics/ClassLength
     confirm_status(:compute_latlng)
     update_column(:status, 'processed')
     delete_tmp_file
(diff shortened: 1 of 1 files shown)


app/models/bgeigie_import.rb
```diff
@@ -97,7 +97,7 @@ def process
     confirm_status(:compute_latlng)
     update_column(:status, 'processed')
     delete_tmp_file
-    update_column(:version, $LAST_MATCH_INFO[1].to_s) if /^# format=(\d+\.\d+\.\d+)$/m =~ source.file.read
+    update_column(:version, $LAST_MATCH_INFO[1].to_s) if /^# format=(.*?)$/m =~ source.file.read
```
matschaffer: I think we can just do a greedy match if we leave out the multi line field. Seems simpler.

ZacharyCChang0828: If I just take the ? out it will grab everything in the file. What else should I do?

auspicacious: I think that if you remove the `m` at the end it will no longer match all lines, only to the end of the current line. And dropping the `?` will make it a greedy match. So:

`/^# format=(.*)$/`

PR #642, opened 2020-03-04
create device story device listing page

eitoball:


matschaffer:
Hey @eitoball, thanks for the work here, but I think I'm going to close this in favor of https://github.com/Safecast/safecastapi/pull/680 

The vue.js approach is interesting but I worry it might be confusing for folks like interns trying to contribute. I appreciate your contributions, as always :)

PR #656, opened 2020-04-21
Fix a default locate switcher

sasharevzin:
When loading a home page for the first time it defaults to `EN` while it should be `JP` because the default is `EN`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,6 +53,10 @@ jobs:
           command: |
             bundle exec rubocop
       - run:
+          name: Run assets precompile
+          command: |
+            bundle exec rake assets:precompile
+      - run:
           name: Run rspec
           command: |
             bundle exec rspec --profile 10 \
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,7 +2,7 @@
 
 module ApplicationHelper
   def link_to_switch_locale
-    if params[:locale] == 'en-US'
+    if I18n.locale == :'en-US'
       link_to 'JA', locale: 'ja'
     else
       link_to 'EN', locale: 'en-US'
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ log/localeapp.yml
 .env
 .envrc
 .direnv
+.byebug_history
 
 .elasticbeanstalk/config.yml
 .elasticbeanstalk/app_versions/


auspicacious:
@sasharevzin Thanks, I will try to look at these later today. Are you in our Slack channels?

sasharevzin:
> @sasharevzin Thanks, I will try to look at these later today. Are you in our Slack channels?

Slack? Nope

.circleci/config.yml
```diff
@@ -52,6 +52,10 @@ jobs:
           name: Run rubocop
           command: |
             bundle exec rubocop
+      - run:
+          name: Run assets precompile
+          command: |
+            bundle exec rake assets:precompile
```
matschaffer, line 58: Can't say I see a huge change in time here comparing https://circleci.com/gh/Safecast/safecastapi/882 to https://circleci.com/gh/Safecast/safecastapi/867

But probably not a bad idea to ensure precompile works as part of the build.

sasharevzin: I compared https://circleci.com/gh/Safecast/safecastapi/879 (Rspec 2:10) and https://circleci.com/gh/Safecast/safecastapi/882 (Rspec 0:51) 
I guess 879 was running without cache and took longer 🤔 
PS. Might be better to use system tests rather testing controller and views ;)

matschaffer: Nice. TIL. This is the only rails app I work on and it's not on 5 yet, so I missed the news. We have a decent set of acceptance specs, but system tests look like they'll be a win when we can switch to them.

app/helpers/application_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module ApplicationHelper
   def link_to_switch_locale
-    if params[:locale] == 'en-US'
+    if I18n.locale == :'en-US'
```
matschaffer, line 5: Yeah this was curious to me to (since @eitoball mentioned it in the other Pr)

Looks like it’s a constant (presumably the system locale) rather than the user selected locale. Is that right?

matschaffer: Seems to work though. https://dev.safecast.org/en-US

The locale selector in this site has always confused me really. Seems like it should be header or session based rather than the route.

sasharevzin: You just basically compare what is the current locale. The problem was that each time user lands to root URL, `params[: locale]` will be `nil` resulting in showing English while default in settings already was set to English https://github.com/Safecast/safecastapi/blob/master/config/application.rb#L33

auspicacious: We had a discussion on a call a while back about inconsistent handling of locale, it really needs more improvement at some point in the future.

That being said, even though this looks like a global constant to me (?), `I18N.locale` is set here: https://github.com/Safecast/safecastapi/blob/master/app/controllers/application_controller.rb#L59 which is also what's recommended in https://guides.rubyonrails.org/v4.2/i18n.html

auspicacious:
@sasharevzin Thanks for these changes.

PR #657, opened 2020-04-21
Translate drive delete link

### LINKED ISSUE #630: Translate drive delete link
@jamoross mentioned this was causing confusion since it's only in english (even when profile is set to JP)

Screenshot_2_7_20__2_21_PM

sasharevzin:
Bgeigie_Import__1____bGeigieデータのインポート____Safecast_API


sasharevzin:
> I always worry about using あなた but I checked with Kaori and she said 👍 so LGTM

My wife is Japanese so I just did copy/paste ;)

PR #658, opened 2020-04-25
Upgrade to Rails 5.2

sasharevzin:
I added some files that come by default with Rails 5. They can be used for future configurations.


app/helpers/application_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module ApplicationHelper
   def link_to_switch_locale
-    if params[:locale] == 'en-US'
+    if I18n.locale == :'en-US'
```
eitoball, line 5: Can we switch language with this?I thought that we should look for `locale` in parameter to switch.

app/helpers/bgeigie_imports_helper.rb
```diff
@@ -18,7 +18,7 @@ def bgeigie_nav_li(status) # rubocop:disable Metrics/AbcSize
                params[:by_status] == status.to_s
              end
     content_tag(:li, class: ('active' if active)) do
-      p = params.except(:action, :controller).merge(by_status: (status unless status == :all))
+      p = params.permit!.except(:action, :controller).merge(by_status: (status unless status == :all))
```
eitoball, line 21: I believe that `permit!` is called to get `Hash`-like object. I think that `#to_h` or `#to_unsafe_h` would be better to get `Hash`-like object.

app/helpers/application_helper.rb
sasharevzin: @eitoball I made this PR based on my master so after merging another PR then we should be good: https://github.com/Safecast/safecastapi/pull/656

app/helpers/bgeigie_imports_helper.rb
sasharevzin: `to_h` won't work because of `unable to convert unpermitted parameters to hash`. I'll try to use `to_unsafe_h`

sasharevzin: @eitoball Ready a6ff470

bin/setup
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env ruby
```
auspicacious, line 1: I'm not sure this script lines up with the way we set up our databases in dev. See https://github.com/Safecast/safecastapi/wiki/Dev:-Setup-on-Linux-and-OS-X-using-Docker

app/helpers/bgeigie_imports_helper.rb
```diff
@@ -18,7 +18,7 @@ def bgeigie_nav_li(status) # rubocop:disable Metrics/AbcSize
                params[:by_status] == status.to_s
              end
     content_tag(:li, class: ('active' if active)) do
-      p = params.except(:action, :controller).merge(by_status: (status unless status == :all))
+      p = params.to_unsafe_h.except(:action, :controller).merge(by_status: (status unless status == :all))
```
auspicacious, line 21: I realize that this will require deeper knowledge of the system, but, for security reasons, most likely we should be calling `.permit` somewhere and `to_h`, right?

https://stackoverflow.com/a/38859674

config/environments/production.rb
```diff
@@ -1,75 +1,98 @@
...
-  # Use a different logger for distributed setups
-  # config.logger = SyslogLogger.new
+  # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
+  # config.force_ssl = true
 
-  # Use a different cache store in production
-  # config.cache_store = :mem_cache_store
+  # Use the lowest log level to ensure availability of diagnostic information
+  # when problems arise.
+  config.log_level = :debug
```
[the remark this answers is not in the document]
sasharevzin: I was thinking it's nice to have more info if something goes wrong after deploying. I can change it to `info`. Please, let me know :)

config/initializers/application_controller_renderer.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
sasharevzin: New initializer configuration for application controller renderer.  This is a really good feature that finally allows dropping some “bypass like” solutions. If you didn’t render outside of controllers scope, you can leave this commented.

docker-sync.yml
```diff
@@ -0,0 +1,17 @@
+version: "2"
```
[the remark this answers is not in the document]
sasharevzin: Correct: http://docker-sync.io/

spec/spec_helper.rb
```diff
@@ -52,13 +52,13 @@
...
+    FileUtils.rm_rf Dir[Rails.root.join('tmp/cache/assets/**/*')]
   end
 
   config.after(:all) do
-    FileUtils.rm_f Rails.root.join('tmp', 'capybara', '*')
-    FileUtils.rm_rf Dir[Rails.root.join('tmp', 'cache', 'assets', '**', '*')]
-    FileUtils.rm_rf Dir[Rails.root.join('public', 'uploads', '**', '*')]
+    FileUtils.rm_f Rails.root.join('tmp/capybara/*')
+    FileUtils.rm_rf Dir[Rails.root.join('tmp/cache/assets/**/*')]
+    FileUtils.rm_rf Dir[Rails.root.join('public/uploads/**/*')]
```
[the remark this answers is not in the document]
sasharevzin: It's just a rubocop thing: https://github.com/rubocop-hq/rubocop-rails/issues/195

bin/setup
sasharevzin: If someone wants to go native then they can use it here but again, this is a default file that comes with Rails 5, and we can delete it

app/helpers/bgeigie_imports_helper.rb
sasharevzin: Right. The goal of this PR is to have Rails 5 running. Later on, we can start refactoring because all these `*_nav_li` methods smell a little 👃 ;)

bin/setup
auspicacious: I think the way we initialize databases might be different enough to cause problems, so we should delete this.

https://github.com/Safecast/safecastapi/blob/master/bin/initialize_databases.bash

config/environments/production.rb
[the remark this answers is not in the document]
auspicacious: Yes, let's start with `info` and then debug if necessary.

config/initializers/application_controller_renderer.rb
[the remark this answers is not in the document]
auspicacious: I don't know enough about Rails to understand this answer, so I will just ask more directly, why would you leave commented-out code in your application?

docker-sync.yml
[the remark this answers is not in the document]
auspicacious: Asking again, will this interfere with people using the setup we already have?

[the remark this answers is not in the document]
matschaffer: It worked fine for me, but I keep ruby outside a container (easier to start the IntelliJ debugger that way)

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+# This file contains migration options to ease your Rails 5.2 upgrade.
+#
+# Once upgraded flip defaults one by one to migrate to the new default.
+#
+# Read the Guide for Upgrading Ruby on Rails for more info on each option.
+
+# Make Active Record use stable #cache_key alongside new #cache_version method.
+# This is needed for recyclable cache keys.
+# Rails.application.config.active_record.cache_versioning = true
```
eitoball, line 12: Should't we uncomment this and other settings in this file?

config/initializers/new_framework_defaults_5_2.rb
```diff
@@ -0,0 +1,39 @@
...
+
+# Store boolean values are in sqlite3 databases as 1 and 0 instead of 't' and
+# 'f' after migrating old data.
+# Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true
+
+# Use SHA-1 instead of MD5 to generate non-sensitive digests, such as the ETag header.
+# Rails.application.config.active_support.use_sha1_digests = true
+
+# Make `form_with` generate id attributes for any generated HTML tags.
+# Rails.application.config.action_view.form_with_generates_ids = true
```
eitoball, line 39: Maybe, append [this content](https://github.com/rails/rails/blob/v5.1.7/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_5_1.rb.tt)?

docker-sync.yml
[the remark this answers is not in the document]
eitoball: I don't think that we should add this. docker-sync is useful for who uses Docker on Mac, but it sometimes stops syncing. Maybe writing down in wiki page; https://github.com/Safecast/safecastapi/wiki/Dev:-Setup-on-Linux-and-OS-X-using-Docker ?

app/safecast/actions/bgeigie_imports/kml.rb
```diff
@@ -17,7 +17,7 @@ def execute(context, filename = nil)
         'app/views/bgeigie_imports/bgeigie_logs.kml.erb'
 
       def send_opts(filename)
-        opts = { type: Mime::KML.to_s }
+        opts = { type: Mime::Type.lookup('kml').to_s }
```
eitoball, line 20: I think that we need MIME type here.

```suggestion
        opts = { type: Mime[:kml].to_s }
```

```
[1] pry(main)> Mime::Type.lookup('kml').to_s
=> "kml"
[2] pry(main)> Mime[:kml].to_s
=> "application/vnd.google-earth.kml+xml"
```

spec/controllers/bgeigie_imports_controller_spec.rb
```diff
@@ -227,11 +227,11 @@
     end
 
     before do
-      get :kml, id: bgeigie_import.id
+      get :kml, params: { id: bgeigie_import.id }
     end
 
     it { expect(response).to be_ok }
-    it { expect(response.content_type).to eq(Mime::KML.to_s) }
+    it { expect(response.content_type).to eq(Mime::Type.lookup('kml').to_s) }
```
eitoball, line 234:
```suggestion
    it { expect(response.content_type).to eq(Mime[:kml].to_s) }
```

spec/acceptance/user_submits_reading_spec.rb
```diff
@@ -17,7 +17,7 @@
     click_button('Submit')
     expect(page).to have_content('123')
     expect(page).to have_content('cpm')
-    expect(user.measurements.size).to eq(1)
```
eitoball, line 20: Did you change to `#count` because `#size` did not return `1`? I am curious.

docker-sync.yml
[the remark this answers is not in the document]
sasharevzin: @eitoball Sure, I'll update in wiki

config/initializers/new_framework_defaults_5_2.rb
sasharevzin: As long as we use any caching

config/initializers/new_framework_defaults_5_2.rb
sasharevzin: @eitoball here the right file for the current version: https://github.com/rails/rails/blob/v5.2.4/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_5_2.rb.tt

spec/acceptance/user_submits_reading_spec.rb
sasharevzin: Just a habit of optimizing the queries. I reverted the change

config/initializers/new_framework_defaults_5_2.rb
eitoball: This file lists new configurations that are added in Rails 5.2. The file that I mentioned is for new defaults in Rails 5.1. If we change `config.load_defaults` in `config/application.rb` to `5.1` or `5.2`, new defaults becomes effective, and application might fail. Therefore, I think that we should append or add new file like `new_framework_defaults_5_1.rb`.

sasharevzin: That’s why we have specs. I tested with rails 5.2 default and it seems working. Let me know if you still think we need to use 5.1 defaults even tho this PR upgrades to 5.2

config/initializers/new_framework_defaults_5_2.rb
eitoball: I was wrong. Setting is this file is new defaults in 5.2. So, we should uncomment and check application before changing `config.load_defaults` to `5.2` in `config/application.rb`

spec/acceptance/user_submits_reading_spec.rb
eitoball: `#count` always issue query. `#size` issues query if relation is not loaded, and uses `#length` if loaded. If `acting` means `loaded`, you are right. `#length` loads relation if not loaded, and use `#length` to count. 

At this line, I don't think `measurements` relation has not been loaded. That is why I was curious to this change. So, does this example fails unless you change this line?

config/initializers/application_controller_renderer.rb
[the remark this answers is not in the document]
eitoball: This file is created when you run `rails app:update` or `rake rails:update`. I think that we should leave as is until we add actual configuration.

docker-sync.yml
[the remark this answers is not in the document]
eitoball: Oh... What I wanted to say is that `we should NOT add this file,  and this content should go into wiki` because you need to use another binary, and some people like running rails outside docker container or running container on Linux .

spec/spec_helper.rb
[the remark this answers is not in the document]
eitoball: Yep.  ruby will convert to path separator depending on OS, so it is okay to use `/` here.

docker-sync.yml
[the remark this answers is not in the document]
sasharevzin: But what to do with these like me who is using docker on his Mac?

config/initializers/new_framework_defaults_5_2.rb
sasharevzin: @eitoball Updated ded3994

spec/acceptance/user_submits_reading_spec.rb
sasharevzin: `measurements` loaded and size will return `size` while after `acting` it was updated in db. and yes, spec fails with `size`.

config/initializers/new_framework_defaults_5_2.rb
eitoball: Yes, we have specs, but specs don't find bugs. specs only make sure what you expect are working in way that you expected.

Setting `config.load_defauls` to `5.2' also set 5.1 defaults, so I think that we should put 5.1 defaults like 5.2 defaults in this file or another separate file.

sasharevzin: > Yes, we have specs, but specs don't find bugs. specs only make sure what you expect are working in way that you expected.

Well, specs let you catch bugs in your logic before code goes out for review/release. They can also, depending on what and how you test, help you make sure your code is working before it's in a deployable state. Part of the reason that we trust specs is that we've seen them fail. It's like [double-entry bookkeeping](https://en.wikipedia.org/wiki/Double-entry_bookkeeping). The specs keep the [SUT](https://en.wikipedia.org/wiki/System_under_test) in place, and the SUT makes the tests pass.

Anyways, here is an update 4e6e446

config/initializers/secret_token.rb
```diff
@@ -6,7 +6,5 @@
 # If you change this key, all old signed cookies will become invalid!
 # Make sure the secret is at least 30 characters and all random,
 # no regular words or you'll be exposed to dictionary attacks.
-Safecast::Application.config.secret_token =
-  '39c00ee18c6a2edd472f8f8b5954b6a332636e32b4c80ed9abc9053cc33bf7f1913d4cb6267ea741d10b48acfd9b3b3df2336fd02de83a4290509bc5d97c0edd'
 Safecast::Application.config.secret_key_base =
-  'f0906ce1f97f1d3a7b1fba927645046ef10d9f7766d2141e023dc074529d9e2b10294e9a6490b04aff9d95ce4764c65e3280a30e1dd2e94f37ba22381ec2b4a4'
+  '39c00ee18c6a2edd472f8f8b5954b6a332636e32b4c80ed9abc9053cc33bf7f1913d4cb6267ea741d10b48acfd9b3b3df2336fd02de83a4290509bc5d97c0edd'
```
auspicacious, line 10: This is a security problem and we should fix it soon after doing the upgrade. I created https://github.com/Safecast/safecastapi/issues/665

config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
```
auspicacious, line 7: I don't really like having inaccurate configuration in the project -- I would assume that a production deployment would fail, because we don't have a Redis instance. However, since we've tested the branch in dev already, I'm going to assume that it works as-is.

Issue #647, opened 2020-04-01
Upgrade API to Rails 6


auspicacious:

sasharevzin:
A pretty big jump, maybe v5 first?

sasharevzin:
Currently blocked by the ruby version. Rails 6 requires at least Ruby 2.5. I see @eitoball is working on it https://github.com/Safecast/safecastapi/pull/669

eitoball:
https://github.com/Safecast/safecastapi/pull/798

PR #669, opened 2020-05-06
Ruby 2.5.8

eitoball:
ruby 2.5.8 is supported on Elastic Beanstalk. (https://docs.aws.amazon.com/elasticbeanstalk/latest/platforms/platforms-supported.html#platforms-supported.ruby) 

The following are blogs related to Ruby 2.5.x updates. I don't think that changes in Ruby 2.5.x don't break api application.

- https://www.ruby-lang.org/en/news/2017/12/25/ruby-2-5-0-released/
  - [NEWS for Ruby 2.5.0](https://github.com/ruby/ruby/blob/v2_5_0/NEWS)
  - Compatibility Issues
    - https://github.com/ruby/ruby/blob/v2_5_0/NEWS#compatibility-issues-excluding-feature-bug-fixes-
    - https://github.com/ruby/ruby/blob/v2_5_0/NEWS#stdlib-compatibility-issues-excluding-feature-bug-fixes-
- https://www.ruby-lang.org/en/news/2018/03/28/ruby-2-5-1-released/
  - `This release includes some bug fixes and some security fixes.`
- https://www.ruby-lang.org/en/news/2018/10/17/ruby-2-5-2-released/
  - `This release includes some bug fixes and some security fixes.`
- https://www.ruby-lang.org/en/news/2018/10/18/ruby-2-5-3-released/
  - `This release is just for fixing the packaging issue. It does not contain any additional bug fixes from 2.5.2.`
- https://www.ruby-lang.org/en/news/2019/03/13/ruby-2-5-4-released/
  - `This release includes bug fixes and a security update of the bundled RubyGems. `
- https://www.ruby-lang.org/en/news/2019/03/15/ruby-2-5-5-released/
  - `This release includes a bug fix for a deadlock in multi-thread/multi-process applications`
- https://www.ruby-lang.org/en/news/2019/08/28/ruby-2-5-6-released/
  - `This release includes about 40 bug fixes after the previous release, and also includes a security fix.`
- https://www.ruby-lang.org/en/news/2019/10/01/ruby-2-5-7-released/
  - `This release includes security fixes...`
- https://www.ruby-lang.org/en/news/2020/03/31/ruby-2-5-8-released/
  - `This release includes security fixes.`


sasharevzin:
@eitoball Let me know what is the status of this PR, please

eitoball:
> Let me know what is the status of this PR, please

I thought that PR was in draft which meaning WIP (work in progress). Anyway, I needed to time to check out changes in Ruby 2.5.x, and check availability in Elastic Beanstalk.

sasharevzin:
> > Let me know what is the status of this PR, please
> 
> I thought that PR was in draft which meaning WIP (work in progress). Anyway, I needed to time to check out changes in Ruby 2.5.x, and check availability in Elastic Beanstalk.

Sure thing. I was poking around to see what needs to be done upgrade to rails 6 (#647) and seems like rails works with Ruby 2.5 >= so was curious about status :)

matschaffer:
https://docs.aws.amazon.com/elasticbeanstalk/latest/platforms/platforms-supported.html#platforms-supported.ruby says 2.5.8

sasharevzin:
> https://docs.aws.amazon.com/elasticbeanstalk/latest/platforms/platforms-supported.html#platforms-supported.ruby says 2.5.8

Thanks @matschaffer 
@eitoball Im curious, shell we skip 2.5 and jump right to 2.6?

matschaffer:
2.6.6 in that case. If it works I’m 👍 we can deploy to dev as a final check

eitoball:
> Im curious, shell we skip 2.5 and jump right to 2.6?

We could, but I wouldn't do. If something happens, it would be easier to investigate because changes are smaller.

sasharevzin:
> > Im curious, shell we skip 2.5 and jump right to 2.6?
> 
> We could, but I wouldn't do. If something happens, it would be easier to investigate because changes are smaller.

Sorry to ask but where smaller changes? In ruby version or in the app?

eitoball:
> Sorry to ask but where smaller changes? In ruby version or in the app?

In ruby. When upgrading ruby version, if something happens, it could be caused by ruby's incompatible changes.

sasharevzin:
> In ruby. When upgrading ruby version, if something happens, it could be caused by ruby's incompatible changes.

But the same thing applies to 2.5, no? I think if there is an issue, we can always rollback to 2.5 from 2.6. It worth trying. Plus 2.6 will have a longer life than 2.5, right?

matschaffer:
I think what he’s getting at is that smaller ruby version jumps will likely mean smaller required code changes (either in the app or gem upgrade). And less risk of getting something wrong or mixed up. 

That said, each jump does require some additional beanstalk env juggling so if the code is easy jumping versions it could make for less aws work.

auspicacious:
It's probably more efficient to upgrade to 2.5, run for ~1 week, then upgrade to 2.6, than to continue to debate. Changing platforms is not too difficult in Elastic Beanstalk. I'll try to do that this week.

sasharevzin:
@auspicacious Thank you for your input (really). 
My last message at this PR. I promise :)
If we have a pretty solid happy path test coverage (~82% at the last run) than maybe we can give a try for 2.6? If it will break something we can always roll it back, write a test to reproduce the failure, fix the failure, and try again? What do you think?

auspicacious:
I am not making a judgement about which opinion about the risk is right. My opinion is that we should be able to deploy and iterate rapidly, even when changing platforms, and that is most important, so that we can accept small contributions from many people.

matschaffer:
Since we have this PR, and it passes, and @auspicacious is willing to work on the upgrade.

I'll say let's go 2.5 first.

@sasharevzin - if you felt inclined to start a 2.6 PR based on this one, I don't see a problem there.

Though @eitoball had been doing ruby/rails/ruby/rails in alternating upgrades. So if there's a rails version that makes sense to bump first maybe we should do that.

Wish I could give more specific advise, but I don't have the time to keep up on what ruby/rails combos are compatible these days.

PR #671, opened 2020-05-06
Use counter_cache

sasharevzin:
`counter_cache` wasn't really used in the right place. Instead of using `#count`, it should be `#size` (or use `#measurements_count` directly) to avoid additional `SQL`
https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache
```
[13] pry(main)> User.first.measurements.count
  User Load (0.8ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1]]
   (9.5ms)  SELECT COUNT(*) FROM "measurements" WHERE "measurements"."user_id" = $1  [["user_id", 1]]
=> 46383
[14] pry(main)> User.first.measurements.size
  User Load (0.5ms)  SELECT  "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> 46383
```
@eitoball I saw you just closed your PR (https://github.com/Safecast/safecastapi/pull/556). Do you have time to review?  Ref #547

### DIFF
--- a/app/views/dashboards/show.html.erb
+++ b/app/views/dashboards/show.html.erb
@@ -40,14 +40,14 @@
       <li>
         <%= link_to [:measurements] do %>
           <em>
-            <%= number_with_delimiter current_user.measurements.count %>
+            <%= number_with_delimiter current_user.measurements.size %>
           </em>
           <%= t('.submissions') %>
         <%- end -%>
       </li>
       <li>
         <%= link_to [:bgeigie_imports] do %>
-          <em><%= current_user.bgeigie_imports.count  %></em>
+          <em><%= current_user.bgeigie_imports.size  %></em>
           <%= t('.bgeigie_imports') -%>
         <%- end -%>
       </li>


eitoball:
I thought that there was reasons to use `count` in these places, but as long as looking at commit logs there seems to be no reason.

I remember that `measurements_count` had totally wrong values, and Mat fixed to run `reset_counter for all users. I guess that inserting measurements directly when importing them causes it even though it uses `reset_counter` [here](https://github.com/Safecast/safecastapi/blob/master/app/models/bgeigie_import.rb#L331-L335) .

sasharevzin:
> I thought that there was reasons to use `count` in these places, but as long as looking at commit logs there seems to be no reason.
> 
> I remember that `measurements_count` had totally wrong values, and Mat fixed to run `reset_counter for all users. I guess that inserting measurements directly when importing them causes it even though it uses `reset_counter` [here](https://github.com/Safecast/safecastapi/blob/master/app/models/bgeigie_import.rb#L331-L335) .

Oh, right, I forgot we do raw insert then, yeah. The counter was wrong but it [should be fixed] (as you mention)(https://github.com/Safecast/safecastapi/blob/master/app/models/bgeigie_import.rb#L137). I guess we are good to merge?

PR #672, opened 2020-05-06
Secure api secret_key_base

sasharevzin:
@auspicacious The only thing that left to do is to set `SECRET_KEY_BASE`. Would you copy form master?


config/secrets.yml
```diff
@@ -0,0 +1,8 @@
+development:
+  secret_key_base: bef4afd7fbae856db0cbdee7f794e7075d9ce25c2e1b9f35439e525aa7fc07c49489fe1e6163a908c9eabfe744783602fb520f04655f78616a8e32c0d7a29fbe
+
+test:
+  secret_key_base: 7b243bc741a9bc9e655a64d038d91e076f8c04f77f32bf3f8e4450a0c4bd59d5490f17a53459353a3c0d5dea2280b6f8d9426495cf46d721e88c94799a274058
+
+production:
+  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
```
eitoball, line 8: @matschaffer Can we set this environment variable in somewhere in Elasticbeanstalk?

auspicacious: @eitoball Yes, we can. We will do so before accepting this PR.

matschaffer: I just did it now. Should be all set.

PR #673, opened 2020-05-06
Try out Falcon server

sasharevzin:
I personally don't see much benefits of using Falcon but here is PR
It works on my local but not sure how to test it on staging. @matschaffer, do you think you can help?
```
app_1            | => Booting Falcon
app_1            | => Rails 5.2.4.2 application starting in development on http://0.0.0.0:3000
app_1            | => Run `rails server -h` for more startup options
```


matschaffer:
Yeah, I think we’d probably have to rewrite a lot of code to see any benefit (if there is any)

I filed this as an idea back during a RubyKaigi partly to get folks who seemed interested in falcon to try contributing to the project. 

Probably not worth really doing.

sasharevzin:
> Yeah, I think we’d probably have to rewrite a lot of code to see any benefit (if there is any)
> 
> I filed this as an idea back during a RubyKaigi partly to get folks who seemed interested in falcon to try contributing to the project.
> 
> Probably not worth really doing.

Agree. So let's close a ticket and PR?

matschaffer:
Makes sense to me. Thanks for trying it out anyway!

ioquatix:
Do you have any use for WebSockets, HTTP/2 or incremental streaming (upstream and/or downstream)?

sasharevzin:
> Do you have any use for WebSockets, HTTP/2 or incremental streaming (upstream and/or downstream)?

Hm... I don't recall...

PR #682, opened 2020-05-13
Remove explicit usage of RAILS_ENV

sasharevzin:
Default `RAILS_ENV` was added to [docker-compose](https://github.com/Safecast/safecastapi/blob/master/docker-compose.yml#L13) file and `rspec` was [running](https://github.com/Safecast/safecastapi/blob/master/spec/spec_helper.rb#L13) in `development` env. I assume this is why `RAILS_ENV` was [injected](https://github.com/Safecast/safecastapi/blob/master/docker_bin/run_test.bash#L3).

I think it makes sense to let `spec_helper.rb` set `RAILS_ENV`. 

What do you think?

### DIFF
--- a/bin/run_tests.bash
+++ b/bin/run_tests.bash
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-docker-compose run --rm --entrypoint bash -c "RAILS_ENV='test' bundle exec rspec spec" app
+docker-compose run --rm --entrypoint bash -c "bundle exec rspec spec" app
--- a/bin/run_tests.bash
+++ b/bin/run_tests.bash
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-docker-compose run app bash -c "bundle exec rspec spec" --rm --entrypoint
+docker-compose run --rm app bash -c "bundle exec rspec spec"
--- a/bin/run_tests.bash
+++ b/bin/run_tests.bash
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-docker-compose run --rm --entrypoint bash -c "bundle exec rspec spec" app
+docker-compose run app bash -c "bundle exec rspec spec" --rm --entrypoint
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,6 @@ services:
     environment:
       DATABASE_HOST: 'postgresql'
       ELASTICSEARCH_URL: 'http://elasticsearch:9200'
-      RAILS_ENV: 'development'
     volumes:
       - .:/src
     depends_on:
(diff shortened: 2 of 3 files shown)


auspicacious:
I'll have to test your changes locally. Historically, even before I started using Docker, I have not been able to run tests without explicitly setting `RAILS_ENV` alongside the `rspec` commnand.

However, when I wasn't using Docker, I similarly had `export RAILS_ENV='development' in my `.bashrc`. I thought it was necessary to set `RAILS_ENV` somewhere, but I can find very little official documentation on Rails behavior.

Maybe you have found the explanation, so I'll give it a try.

sasharevzin:
> Maybe you have found the explanation, so I'll give it a try.

Right, when I was running rspec I was confused of env, like why it starts from development :(
Let me know if this explanation is good enough to remove this env from the container :)

bin/run_tests.bash
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-docker-compose run --rm --entrypoint bash -c "RAILS_ENV='test' bundle exec rspec spec" app
+docker-compose run --rm --entrypoint bash -c "bundle exec rspec spec" app
```
auspicacious, line 3:
```suggestion
docker-compose run --rm --entrypoint 'bash -c "bundle exec rspec spec"' app
```

When I tried to run the script, I had to quote the entire entry point before it worked -- I'm not sure why it used to work.

Otherwise, looks good.

sasharevzin: Thanks, sorry for not running this script. I rearranged params and not it works. Would you take a look again, please?

bin/run_tests.bash
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-docker-compose run --rm --entrypoint bash -c "RAILS_ENV='test' bundle exec rspec spec" app
+docker-compose run app bash -c "bundle exec rspec spec" --rm --entrypoint
```
auspicacious, line 3: No, I think something changed in the CLI parser, because it used to work.

You changed my suggestion, and your new code works, but:

* The documentation says that `--entrypoint` requires an argument. In your command line, I think that `--entrypoint` is being ignored. I don't think it's necessary to use `--entrypoint`.
* The documentation says that options `--rm` and `--entrypoint` should be written after `run` but before `SERVICE`, which is probably why I had used the ordering that I did.

In case Docker fixes the parser, best to use the order that is specified in the documentation:

https://docs.docker.com/compose/reference/run/

```suggestion
docker-compose run --rm app bash -c "bundle exec rspec spec"
```

sasharevzin: Great thanks for checking for me :)

auspicacious:
Thank you for the change, I retested and it looks good.

PR #686, opened 2020-05-24
Automate deployment

auspicacious:
Opening pull request to start testing builds.


auspicacious:
I'm going to move the deployment script into the infrastructure repository, but there is one small change to the `package.py` script that needs to be made.

As I understand it, when you upload a version to EB, AWS marks it as having a particular state associated with the AWS CI solution. Because we were setting it to "--process," AWS was trying to process the package on upload (checking if the upload is valid somehow?), and always failing for some reason, making the status of the version `FAILED`. The AWS GUI will happily let you deploy a failed package, but the AWS API will not let you deploy it.

This change leaves the version in state `UNPROCESSED`, which can be deployed through the API.

Issue #648, opened 2020-04-01
Customize api nginx to rate-limit bots


auspicacious:
We should both:

* Set up nginx rate-limiting to automatically slow down crawlers, bots, etc.
* Create a pattern to allow us to explicitly block IPs or IP address ranges

This should be accomplished through .ebextensions.

sasharevzin:
I would suggest using middleware https://github.com/kickstarter/rack-attack
PS. I understand that it's better to do it via Nginx, so the request won't even ping the app, but rack-attack is way more convenient with all its options, and it's a middleware, so that rails won't be loaded. Currently using for https://streeteasy.com/ and it works great

auspicacious:
It's interesting code, but I feel like nginx is a more general solution that we can re-use in other places in the future easily. Plus it's pretty easy to do with Elastic Beanstalk.

auspicacious:
I thought I would restate our priorities a little:

These two will require the nginx rate limiting module:
* Automatically rate limit IP addresses that are making too many requests too quickly
* Allow certain IP addresses to be whitelisted to make more requests without limiting (e.g. https://realtime.safecast.org/)

These two can be accomplished with more basic nginx configuration:
* Make it easy to completely block specific IP addresses
* Make it easy to completely block specific user-agents (this has been useful in the past)

PR #700, opened 2020-06-17
Nginx docker temp

ZacharyCChang0828:
I moved the entire nginx configuration file into the repository and modified docker-compose so it now use it as it's volume.
blockips.conf is added to specify IPs to block. 
1_webapp.conf contains rate limiting and configuration to connect nginx to the app.


auspicacious:
@ZacharyCChang0828 Thank you, I will try to look through it this afternoon.

Related to issue https://github.com/Safecast/safecastapi/issues/648

nginx/conf.d/1_webapp.conf
```diff
@@ -0,0 +1,47 @@
+#Modified from: https://itnext.io/docker-rails-puma-nginx-postgres-999cd8866b18
```
auspicacious, line 1: About the filename: files are sorted in lexical order, not numeric order, so this will not behave the way that you think it will.

```
01_file
02_file
09_file
10_file
1_file
11_file
50_file
99_file
```

The convention is to use two digits, and if the ordering doesn't matter, use `50_` to put it in the middle.

nginx/conf.d/rate_limit.conf
```diff
@@ -0,0 +1 @@
+
```
auspicacious, line 1: What's supposed to be here?

nginx/conf.d/proxy.conf.bak
```diff
@@ -0,0 +1 @@
+client_max_body_size 10M;
```
auspicacious, line 1: You can delete the `.bak` files to reduce the distraction.

nginx/conf.d/1_webapp.conf
auspicacious: It looks like the other files aren't following the normal convention, so I guess it doesn't matter that much. `1_` will sort before anything that begins with a letter of the alphabet.

nginx/conf.d/1_webapp.conf
```diff
@@ -0,0 +1,47 @@
...
+
+upstream safecastapi {
```
auspicacious, line 3: I would appreciate it if you could leave a comment on this pull request to describe where you made changes and what changes you made.

nginx/conf.d/1_webapp.conf
```diff
@@ -0,0 +1,47 @@
...
+   server app:3000;
+}
+
+limit_req_zone $binary_remote_addr zone=one:10m rate=10r/s;
```
auspicacious, line 7: This looks like a good start, but, if it is possible, it would be best if we could add this configuration to a new file instead of modifying an existing one.

matschaffer:
Closing out since this has been draft for quite some time and both folks involved aren't active at the monent.

PR #701, opened 2020-06-21
Added git tagging of builds.

auspicacious:
If we do this, it will become much easier to generate diffs of what has changed between deploys.

We will need to grant CircleCI write privs on our repositories, however.


auspicacious:
Closing for now based on security concerns; I have an alternate approach in mind.

PR #707, opened 2020-07-03
Removed store_prefix from store_dir

### LINKED ISSUE #675: Remove store_prefix from store_dir
Following the PR https://github.com/Safecast/safecastapi/pull/587/files#r338905415, I would like to propose removing `store_prefix` (https://github.com/Safecast/safecastapi/blob/master/app/uploaders/file_uploader.rb#L13) on `store_dir` to avoid adding more personal folders to https://github.com/Safecast/safecastapi/blob/master/.gitignore#L13-L15

I don't see any reasons why would we have `store_prefix`

rdeskins:
Fixes #675

### DIFF
--- a/app/uploaders/file_uploader.rb
+++ b/app/uploaders/file_uploader.rb
@@ -7,6 +7,12 @@ class FileUploader < CarrierWave::Uploader::Base
   # include CarrierWave::RMagick
   # include CarrierWave::MiniMagick
 
+   # Override the directory where uploaded files will be stored.
+  # This is a sensible default for uploaders that are meant to be mounted:
+  def store_dir
+    "uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
+  end
+
   def store_prefix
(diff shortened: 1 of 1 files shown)


app/uploaders/file_uploader.rb
```diff
@@ -7,12 +7,6 @@ class FileUploader < CarrierWave::Uploader::Base
   # include CarrierWave::RMagick
   # include CarrierWave::MiniMagick
 
-  # Override the directory where uploaded files will be stored.
-  # This is a sensible default for uploaders that are meant to be mounted:
-  def store_dir
-    "#{store_prefix}uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
-  end
-
   def store_prefix
```
sasharevzin, line 10: Thanks @rdeskins but I think what we really what to do is to get rid of store prefix

matschaffer:
Congrats on the first merge @rdeskins !

PR #708, opened 2020-07-07
Added creation of 2 non-moderator users to rake task

rdeskins:


sasharevzin:
Thanks, @rdeskins. Would you link an issue to this PR so I can get a little bit more details about the motivation for this PR was before reviewing?

rdeskins:
I'm not sure if there's an issue for this yet. In the slack chat, me and @matschaffer were discussing adding non-moderator users and he suggested making a PR for this.

lib/tasks/boostrap.rake
```diff
@@ -21,4 +23,20 @@ namespace :db do
     u.save
     puts "Created user #{u.email} with password #{u.password}"
   end
+
+  desc 'Create non-moderator'
+  task bootstrap: %i(protect environment) do
```
sasharevzin, line 28: We might need to change the task name because the same name was used for `Create dummy user`. 
I would suggest that admin and non-admin users will be created in the same `bootstrap` task

lib/tasks/boostrap.rake
```diff
@@ -10,7 +10,7 @@ namespace :db do
     end
   end
 
-  desc 'Create dummy users'
+  desc 'Create dummy user'
```
sasharevzin: I actually liked this title

rdeskins: I was getting a bunch of failed tests and wasn't sure how to fix it, but changing it back made it work.

sasharevzin: If it fails next times, let me know and we will fix together ;)

spec/fabricators/device_story_fabricator.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Fabricator(:device_story) do
-  device_urn { Random.rand(20..100).to_s }
+  sequence(:device_urn) { |i| "fabricated:#{i}" }
```
sasharevzin, line 4:
```suggestion
  device_urn {  sequence(:device_urn) { |i| "fabricated:#{i}" } }
```

PR #711, opened 2020-07-09
Moved approved/rejected information from bgeigie listing to show page

### LINKED ISSUE #536: Move approved/rejected columns from bgeigie listing to bgeigie show page
Doesn't really need to be on this view

https://api.safecast.org/en-US/bgeigie_imports?by_status=done&locale=en-US&order=created_at+desc

But would be helpful (for moderators only) to see on https://api.safecast.org/en-US/bgeigie_imports/40802

rdeskins:
fixes #536 
Removed approved by/rejected by columns from table on bgeigie listing page. Approved/rejected information is on show page and is only viewable by moderators and uploader.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -5,3 +5,10 @@
  *= require application/safecast
  *= require bootstrap-datetimepicker.min
 */
+
+span {
+    color:#202020;
+    font-size: 13px;
+    font-family: proxima-nova, 
+    sans-serif;
+}
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -6,9 +6,8 @@
  *= require bootstrap-datetimepicker.min
 */
 
-span {
-    color:#202020;
-    font-size: 13px;
-    font-family: proxima-nova, 
-    sans-serif;
+.status-text {
+  color:#202020;
+  font-size: 13px;
+  font-family: proxima-nova, sans-serif;
 }
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -7,7 +7,7 @@
 */
 
 .status-text {
-  color:#202020;
+  color: #202020;
   font-size: 13px;
   font-family: proxima-nova, sans-serif;
 }
--- a/app/helpers/bgeigie_imports_helper.rb
+++ b/app/helpers/bgeigie_imports_helper.rb
@@ -46,8 +46,8 @@ module BgeigieImportsHelper
     user.try!(:moderator?)
   end
 
-  def approval_participant?(user, bgeigie_import)
-    user.moderator? || user == bgeigie_import.user
+  def approval_participant?(bgeigie_import)
+    current_user.moderator? || current_user == bgeigie_import.user
   end
 
   def tilemap_link(bgeigie_import)
--- a/app/helpers/bgeigie_imports_helper.rb
+++ b/app/helpers/bgeigie_imports_helper.rb
@@ -46,6 +46,10 @@ module BgeigieImportsHelper
     user.try!(:moderator?)
   end
 
+  def approval_participant?(user, bgeigie_import)
+    user.moderator? || user == bgeigie_import.user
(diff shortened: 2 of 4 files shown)


app/views/bgeigie_imports/_status.html.erb
```diff
@@ -11,13 +11,20 @@
...
-  </span><span style="color: #202020;font-size: 13px;font-family: proxima-nova, sans-serif;">
-  <b>comment:</b>
-  <%= @bgeigie_import.comment if @bgeigie_import.comment.present? %>
-</span>
+<%- if user_signed_in? && (current_user.moderator? || current_user == @bgeigie_import.user)-%>
+  <%- if @bgeigie_import.rejected? -%>
+    <span class="label label-warning">
+      Rejected by: <%= @bgeigie_import.rejected_by %>
+    </span>
+    <span style="color: #202020;font-size: 13px;font-family: proxima-nova, sans-serif;">
```
matschaffer, line 19: Might be a good time to move this styling into CSS. Mind doing that?

app/views/bgeigie_imports/_status.html.erb
```diff
@@ -11,13 +11,20 @@
...
   </span>
 <%- end -%>
-<%- if @bgeigie_import.rejected == true -%>
-  <span class="label label-warning">
-    Rejected by: <%= @bgeigie_import.rejected_by %>
-  </span><span style="color: #202020;font-size: 13px;font-family: proxima-nova, sans-serif;">
-  <b>comment:</b>
-  <%= @bgeigie_import.comment if @bgeigie_import.comment.present? %>
-</span>
+<%- if user_signed_in? && (current_user.moderator? || current_user == @bgeigie_import.user)-%>
```
matschaffer, line 14: Maybe not for this PR, but would be good to turn this pattern into a helper function. May something like `approval_participant?(current_user, @bgeigie_import)`

Feel free to implement it if you like that idea.

rdeskins:
I wasn't sure where to put the CSS, so I put it in application.css for now. Let me know if there's a better place. 
I also wrote the helper method, but didn't push that yet. I can make a new PR for that one or put it in this one.

app/assets/stylesheets/application.css
```diff
@@ -5,3 +5,10 @@
  *= require application/safecast
  *= require bootstrap-datetimepicker.min
 */
+
+span {
```
sasharevzin, line 9: Having style on the entire tag might be too risky because it can override other styles. let's make it a CSS class instead like we do [here](https://github.com/Safecast/safecastapi/pull/711/files#diff-4ebb75a2f2d31b2591e792ee7b724436R25)

app/views/bgeigie_imports/_status.html.erb
```diff
@@ -1,3 +1,4 @@
+<%= stylesheet_link_tag "application.css" %>
```
sasharevzin, line 1: I think we already [doing](https://github.com/Safecast/safecastapi/blob/master/app/views/layouts/application.html.erb#L13) it

rdeskins:
Yeah, I can put that in this one too since I already made the helper method.

app/helpers/bgeigie_imports_helper.rb
```diff
@@ -46,6 +46,10 @@ def moderator?(user)
     user.try!(:moderator?)
   end
 
+  def approval_participant?(user, bgeigie_import)
+    user.moderator? || user == bgeigie_import.user
+  end
```
sasharevzin, line 51: I think we can just use `current_user` helper inside
```suggestion
  def approval_participant?(bgeigie_import)
    current_user.moderator? || current_user == bgeigie_import.user
  end
```

app/assets/stylesheets/application.css
```diff
@@ -5,3 +5,9 @@
  *= require application/safecast
  *= require bootstrap-datetimepicker.min
 */
+
+.status-text {
+  color:#202020;
```
matschaffer, line 10:
```suggestion
  color: #202020;
```

PR #712, opened 2020-07-12
Remove duplicated option settings in .ebextensions

eitoball:
I would like to remove this because the same option setting (`ConnectionSettingIdleTimeout`) is in `.ebextensions/web.config`.


auspicacious:
Thanks, this looks like a good catch. Is it ready to merge?

eitoball:
> Thanks, this looks like a good catch. Is it ready to merge?

@auspicacious Thank for checking. Yes, it is ready.

PR #717, opened 2020-07-16
Upgrade ruby to 2.6.6

rdeskins:
Discussed upgrading ruby to 2.6.6 before 2.7.1 in #716 

Fixes https://github.com/Safecast/safecastapi/issues/718


Dockerfile
```diff
@@ -16,6 +16,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /src
 ADD Gemfile Gemfile.lock .ruby-version requirements.txt /src/
 RUN LC_ALL=en_US.UTF-8 pip3 install -r requirements.txt
+RUN gem install bundler
```
eitoball, line 19: Did you need this? ruby 2.6 should contain bundler by default.

rdeskins:
Only thing I can think of to change about this is something to do with this line. If we keep it, it would have to be removed eventually, probably when upgrading to 2.7.1 in the future. https://github.com/Safecast/safecastapi/blob/fb0ad507dd1bc4b2b3a7619cccff9e2f92975435/Dockerfile#L19

auspicacious:
@rdeskins If things will still work without `gem install bundler` there, probably best to remove it.

PR #725, opened 2020-08-17
Add Device Story show page

rdeskins:
Here's what I have right now. Tried to have a similar layout to the realtime website, but had to put the map below to make everything fit.
image

### DIFF
--- a/app/views/device_stories/show.html.erb
+++ b/app/views/device_stories/show.html.erb
@@ -17,16 +17,12 @@
                     <td style="font-size: 25px"><%= @device_story.last_values %></td>
                 </tr>
             </table>
-            <a href="http://grafana-dev.safecast.cc/d/DFSxrOLWk/safecast-device-details?orgId=1&from=now-90d&to=now&refresh=15m&var-device_urn=<%= @device_story.device_urn %>"
-                target="_blank";
-                style="font-size: 18px">
-                More sensor data
-            </a>
+            <h4><%= link_to 'More sensor data', grafana_more_data(@device_story.device_urn), :_blank %></h4>
         </div>
 
         <div class="span7">
             <iframe 
-                src="https://grafana-dev.safecast.cc/d-solo/DFSxrOLWk/safecast-device-details?orgId=1&from=now-90d&to=now&refresh=15m&var-device_urn=<%= @device_story.device_urn %>&panelId=14" 
+                src="<%= grafana_iframe(@device_story.device_urn, grafana_panel(:cpm)) %>"
                 width="100%" height="250" frameborder="0">
             </iframe>
         </div>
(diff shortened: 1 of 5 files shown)


robouden:
Good progress!  
Can we have the same scaling/time selection as in Garfana?
Also possible to export the data or links for exporting the displayed data?

Rob

rdeskins:
> Good progress!
> Can we have the same scaling/time selection as in Garfana?
> Also possible to export the data or links for exporting the displayed data?
> 
> Rob

For the scaling/time, do you mean how Grafana will show the last 2 days? I can change the CPM graph to that scaling. 
I'm not sure what you mean by the exporting part though.

robouden:
"For the scaling/time, do you mean how Grafana will show the last 2 days" meaning select the time for the display. Not only 2 days.

For the export see 
Screenshot from 2020-08-18 09-03-23

rdeskins:
With the Grafana iframes, i haven't found a way to add a time range selector when embedding a graph.
I see what you mean now about the exporting. I could add a link to that, but maybe it would make sense to add a link to the whole grafana page, especially since I can't find a way to add time scaling.

robouden:
Robin,

Yes, I came across the same issue with Iframes on RT a long time ago and decided to use Worpress with links to Grafana to make things work. I think it is a necessity to use the time scaling for displaying the data.

rob

matschaffer:
Let's not implement time scrolling on this page. Just provide a link to the grafana view if people want to dig deeper. 

The goal is to replace https://realtime.safecast.org/sensors/200041/ which doesn't include on-page time selection either.

I would just mimic that page for now.

Here's a rough sketch of how I'd lay it out.

USA__NY__Shrub_Oak___SAFECAST

We can add fields to the `device_stories` table for stuff ttserve doesn't provide, though it comes with the assumption that some moderator (or eventually device owner) will help fill those in.

rdeskins:
This is what it looks like right now after after those changes. Imo the table still looks a bit weird, but its easier to read the measurements/dates and notice the grafana link at least.
image

app/views/device_stories/show.html.erb
```diff
@@ -0,0 +1,45 @@
...
+        <div class="span2">
+            <table class="table table-hover">
+                <thead>
+                    <%= time_ago_in_words(@device_story.last_seen) %> ago (<%= @device_story.last_seen %>)
+                </thead>
+                <tr>
+                    <td style="font-size: 25px"><%= @device_story.last_values %></td>
+                </tr>
+            </table>
+            <a href="http://grafana-dev.safecast.cc/d/DFSxrOLWk/safecast-device-details?orgId=1&from=now-90d&to=now&refresh=15m&var-device_urn=<%= @device_story.device_urn %>"
```
matschaffer, line 20: We should swap this out for a configurable value, and set the default to https://grafana.safecast.cc/d/DFSxrOLWk

Nakushi64:
Screenshot of the current state of the layout: Screenshot 2020-11-10 at 14 40 04

matschaffer (approved):

PR #729, opened 2020-09-06
add index to measurements table on updated_at column

### LINKED ISSUE #666: Add updated_at index to measurements table
I was digging around and found that slowest requests are on `MeasurementsController#index` and it happens when someone is trying to fetch the list of Measurements using `since` filter like [/measurements.json](http://api.safecast.org/measurements.json?per_page=1000&since=2020-05-05%2B01%3A15%3A03)

[NewRelic](https://rpm.newrelic.com/accounts/1057812/applications/10580222/datastores#/overview/Postgres/trace/?id=a6b6e952-8e6c-11ea-a7a1-0242ac110007_0_3127&metric=Datastore%252Fstatement%252FPostgres%252FMeasurement%252Ffind) and [Kibana](https://5bc3d4f7330a4459881197a85659caf5.us-west-2.aws.found.io:9243/app/apm#/services/safecastapi-prd/transactions/view?rangeFrom=now-24h&rangeTo=now&refreshPaused=true&refreshInterval=0&traceId=994b74af1b48a758592db089a9ef6f38&transactionId=6c30783bca3d6dd4&transactionName=MeasurementsController%23index&transactionType=request&detailTab=metadata)

Proposed solution is to add an index on `updated_at` OR add another field that will contain only `date` to reduce an index size that won't include time (who care about, right?)

PS. I know, this is a pretty huge table and adding another index might overkill but table has index on `captured_at` so maybe it's ok?

cgeb:
Fixes #666

When I created the migration file and ran it, it made a bunch of unrelated changes to the `structure.sql` file. How does everyone usually make changes to this file? Is it done manually?

When I run this command I get a different version of `pg_dump` then that of the current `structure.sql` (9.5.19). Maybe that is the issue?
```
$ docker-compose run app bash -c "pg_dump --version"
Starting safecastapi_elasticsearch_1           ... done
Starting d1abed9afdee_safecastapi_postgresql_1 ... done
pg_dump (PostgreSQL) 11.8 (Debian 11.8-1.pgdg100+1)
```


matschaffer:
Yeah, I tend to run migrations then do edits. This looks better than usual though so maybe we can use it as is. 

We should also probably try going back to schema.rb at some point. We stopped because rails didn’t handle postgis correctly but I think it does now.

matschaffer (approved):
Gonna go ahead with this in hopes that it helps with the db pressure we get from https://github.com/radoondas/safecastbeat

PR #733, opened 2020-11-13
Linkable search for the Device Story page

Nakushi64:
- Refer to #606.
- Added search functionality via the "LIKE"-query.
- Added Elasticseach functionality which probably just works on my local database.

### DIFF
--- a/app/controllers/device_stories_controller.rb
+++ b/app/controllers/device_stories_controller.rb
@@ -24,6 +24,6 @@ class DeviceStoriesController < ApplicationController
   end
 
   def get_elastic_searched_table(search_term)
-    apply_scopes(DeviceStory).search("*"+search_term+"*").records.page(params[:page]).per(params[:per_page])
+    apply_scopes(DeviceStory).search('*' + search_term + '*').records.page(params[:page]).per(params[:per_page])
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/device_stories_controller.rb
```diff
@@ -4,13 +4,26 @@ class DeviceStoriesController < ApplicationController
...
+                      end
   end
 
   def show
     @device_story = DeviceStory.find(params[:id])
     respond_with @device_story
   end
+
+  def get_like_searched_table(search_term)
+    apply_scopes(DeviceStory).where('lower(device_urn) LIKE :search OR lower(custodian_name) LIKE :search', search: "%#{search_term}%")
```
matschaffer, line 21: I wouldn't expect you'd need both `LIKE` and elasticsearch here.

Once the data is in elasticsearch you should be able to search there. Plus text fields get analyized automatically so no need for cast/lower/etc

matschaffer: Screen Shot 2020-11-13 at 6 58 36

Seems to be working fine locally. So it should show up on prod once deployed as well.

We _might_ need to update the permissions to post, but that shouldn't be a problem.

matschaffer (approved):
Screen Shot 2020-11-13 at 7 23 32

Looks great!

Might be good to start including some tests with these PRs of yours soon. Otherwise we might break stuff down the road.

But I'm happy to have this contribution as is.

PR #734, opened 2020-11-17
Dynamic search

Nakushi64:
Refer to #606 and #733 .

Current issues:
- ~~The search is getting slower and buggier the longer the input is. Caused by querying the database for every keyup event.~~
- ~~Queries that are looking for shorter strings need more time and are displayed later which sometimes leads to wrong search results.~~

Solved with "debouncing".


app/assets/javascripts/application.js
```diff
@@ -20,3 +20,13 @@ jQuery.ajaxSetup({
     xhr.setRequestHeader("Accept", "text/javascript");
   }
 });
+
+$("#search_form").keyup(delay(function() { $('form').find('input').submit()}, 500));
```
matschaffer, line 24: Just noticed we include underscore.js already, so this could use https://underscorejs.org/#debounce rather than defining this delay function.

I found an example at https://jsfiddle.net/missinglink/19e2r2we/

app/views/device_stories/index.js.erb
```diff
@@ -0,0 +1,3 @@
+$('#results_class').html("<%= j (render 'results', device_stories: @device_stories) %>")
```
matschaffer, line 1: I don't think you need the parentheses here. If you want to keep them `j render('results', device_stories: @device_stories)` might read a little more nicely

app/views/device_stories/_results.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <thead>
+    <%= table_sort_header(:device, :device_urn) %>
+    <%= table_sort_header(:device, :custodian_name) %>
+    <%= table_sort_header(:device, :last_values) %>
+    <%= table_sort_header(:device, :last_location_name) %>
+    <%= table_sort_header(:device, :last_seen) %>
+  </thead>
+  <tbody>
+  <% device_stories.each do |device_story| %>
+    <tr onclick = "javascript: location.href = '<%=device_story_path(device_story) %>'">
```
matschaffer, line 11: We have an existing pattern for this using bootstrap rowlink https://github.com/Safecast/safecastapi/blob/5b7d6595e81d2a8c546d7cb66c1a14b70f43ac82/app/views/bgeigie_imports/index.html.erb#L52

Would be good to do the same here.

This also has me wondering if we _really_ want that.

I find myself sometimes navigating when I wanted to just copy/paste some data from the site.

https://www.robertcooper.me/table-row-links#should-table-rows-be-links-in-the-first-place does a good job calling out some of the downsides.

Wonder if @Azby or @nokton opinions on if we should keep all table rows linkable or not.

matschaffer:
This is up on https://dev.safecast.org/en-US/device_stories now - I left some comments in slack regarding pushState to preserve the linkability of the search like we have today on https://api.safecast.org/en-US/device_stories?search=mat

matschaffer (approved):
looks great on https://dev.safecast.org/en-US/device_stories I'd say we can roll with this and adjust whatever we need later on.

PR #737, opened 2020-11-24
Adjust table width

Nakushi64:
Added a button to toggle between layouts for the device story table.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -23,5 +23,5 @@ jQuery.ajaxSetup({
 
 $("#search_form").keyup(_.debounce(function() {
     $('form').find('input').submit();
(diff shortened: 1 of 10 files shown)


matschaffer:
Here's a screenshot for @robouden 

Screen Shot 2020-11-25 at 1 35 45

Seems a little weird to me, but if it helps I don't see a problem merging it.

Would be great to work on a responsive layout at some point after we get comments & attachments working for device stories.

robouden:
Maybe make the width of the last seen filed double size, so we can post much more line at one. If possible even do not use the paging of the data.

robouden (commented):
Seem ok!!

robouden:
approved

matschaffer (approved):

PR #738, opened 2020-11-24
Narrow down search triggering keys

Nakushi64:
Only printable characters and Backspace are triggering the search now.


matschaffer (commented):
Looks like this doesn't trigger on copy-paste (at least not via chrome on macos)

matschaffer (changes requested):
I think https://github.com/Nakushi64/safecastapi/pull/1 might be better

matschaffer (approved):
Works for me, of course ;)

PR #743, opened 2020-12-04
Device story comments

Nakushi64:
Refer to #606 
- Can leave comments on device story pages if signed in
- Can delete own comments
- Moderators can delete everyone's comments


app/models/device_story.rb
```diff
@@ -3,4 +3,6 @@
 class DeviceStory < ApplicationRecord
   validates :device_urn, uniqueness: true
   validates :device_urn, :custodian_name, presence: true
+  has_many  :device_story_comments
+  accepts_nested_attributes_for :device_story_comments
```
matschaffer, line 7: This doesn't seem like the right move to me. I'd probably opt to have the device story comment be its own controller (a nested route could be good so you always have the device story id) and write directly.

Could be that's my old-school rails background talking though. Maybe @sasharevzin or @eitoball would be able to cross check me.

matschaffer:
We should also plan to get https://akismet.com/ in here, otherwise we'll likely start getting spam submissions at some point.

The authenticated guard helps a bit, but some bots do work out registration in order to sneak in spam links.

matschaffer:
We actually had a case not too long ago where a spam bot used the registration process itself since we display the name on pages like https://api.safecast.org/en-US/users/1154 leading to https://github.com/Safecast/safecastapi/pull/631 as a workaround.

app/models/device_story.rb
Nakushi64: Ok, will work on separating them.

matschaffer:
Finally this is big enough that I think we should add a basic acceptance spec. We have examples in https://github.com/Safecast/safecastapi/tree/master/spec/acceptance 

There's also https://github.com/Safecast/safecastapi/blob/master/spec/controllers/device_stories_controller_spec.rb and https://github.com/Safecast/safecastapi/blob/master/spec/safecast/tasks/device_stories/update_metadata_spec.rb which should help provide some background material for a commenting spec.

Nakushi64:
Got it! Will try to implement your suggestions.

matschaffer:
ah, here's the gem I found last time I was googling https://github.com/Safecast/safecastapi/issues/606#issuecomment-592321540

Looks like it should support a basic string check agains the free akismet API. Register your own email for starters, if it works I can handle registering a key for the the deployed sites.

app/models/device_story.rb
eitoball: I agree with @matschaffer. We should use [Nested resources](https://guides.rubyonrails.org/routing.html#nested-resources).

db/migrate/20201204004647_create_device_story_comments.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CreateDeviceStoryComments < ActiveRecord::Migration[5.2]
+  def change
+    create_table :device_story_comments do |t|
+      t.references :device_story, foreign_key: true
+      t.references :user, foreign_key: true
+      t.text :text
```
eitoball, line 8: You might want to have a validation to check size fo `text` in `DeviceStoryComment` because PostgresSQL accepts up to 1 GB (https://wiki.postgresql.org/wiki/FAQ#What_is_the_maximum_size_for_a_row.2C_a_table.2C_and_a_database.3F). I don't think that we would want comments in such size.

Nakushi64:
The comments now have their own controller and I used [this](https://github.com/joshfrench/rakismet) to implement Akismet. Also I limited the size of each comment in the model to 1000 characters. Let me know if there is still anything I should do differently.

app/controllers/device_story_comments_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  private
+
+  def set_device_story_comment
+    @device_story_comment = @device_story.device_story_comments.find(params[:id])
+  end
+
+  def device_story_comment_params
+    params.inspect
```
matschaffer, line 53: guessing this line isn't supposed to be here

app/controllers/device_story_comments_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    @device_story_comment = @device_story.device_story_comments.build
+  end
+
+  def create
+    @device_story_comment = @device_story.device_story_comments.build(device_story_comment_params)
+    respond_to do |format|
+      if @device_story_comment.save && !@device_story_comment.spam?
+        format.html { redirect_to device_story_path(@device_story), notice: 'Comment successfully submitted!' }
+      else
+        flash[:error] = 'Could not submit your comment. Please, try again!'
```
matschaffer, line 21: Would trying definitely help? Might be good to capture the error that surfaced and show something to the user based on that. Of course if it's spam being obscure makes some sense.

config/initializers/rakismet.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+Safecast::Application.config.rakismet.key = '07b3d56e9704'
```
matschaffer, line 3:
```suggestion
Safecast::Application.config.rakismet.key = ENV['AKISMET_API_KEY']
```

config/initializers/rakismet.rb
```diff
@@ -0,0 +1,4 @@
...
+Safecast::Application.config.rakismet.url = 'http://safecast.com/'
```
matschaffer, line 4:
```suggestion
Safecast::Application.config.rakismet.url = "https://#{ENV['DEFAULT_HOSTNAME']}/"
```

app/controllers/device_story_comments_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+        format.json { render :show, status: :ok, location: @device_story_comment }
+      else
+        format.html { render :edit }
+        format.json { render json: @device_story_comment.errors, status: :unprocessable_entity }
+      end
+    end
+  end
+
+  def destroy
+    @device_story_comment.destroy
```
matschaffer, line 40: Pretty sure at the moment this allows any user to delete (or update) any comment, we should set it so only moderator or comment author can do that.

app/views/device_stories/show.html.erb
```diff
@@ -47,5 +49,40 @@
...
+    <%- if @device_story.device_story_comments.empty? %>
+      No Comments yet.
+    <%- else %>
+      <%- @device_story.device_story_comments.each do |comment| %>
+        <tr>
+          <td>
+            <b><%= link_to comment.user.name , comment.user %></b> <div style="display: inline-block; color: dimgrey; font-size: small">(<%= time_ago_in_words(comment.created_at) %> ago)</div> <br>
+            <%= comment.content %>
+            <%- if user_signed_in? -%>
+              <%- if (comment.user_id == current_user.id) || moderator?(current_user) %>
```
matschaffer, line 65: Looks like you got the right idea here, but we should protect the route as well. People could build their own requests to delete the user if the API allows it.

app/views/device_stories/show.html.erb
```diff
@@ -47,5 +49,40 @@
...
+    <%- end %>
+    </tbody>
+  </table>
+</div>
+<div style="margin-left: 5px;min-width: 500px">
+  <%- if user_signed_in? %>
+    <%= form_with model: DeviceStoryComment, url: device_story_device_story_comments_path(@device_story), local: true, method: 'post' do |form| %>
+      <%= form.label :content, current_user.name %>
+      <%= form.text_area :content, :rows => 3, style: 'width:97%;' %>
+      <%= form.hidden_field :user_id, :value => current_user.id %>
```
matschaffer, line 81: We shouldn't have this here since anyone could override it. Move it to `DeviceStoryCommentsController#create`

spec/requests/device_story_comments_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe DeviceStoryCommentsController, type: :request do
+  let!(:device_story) { Fabricate(:device_story, last_location_name: 'Israel', device_urn: 'safecast:114699387') }
+  let!(:user) { Fabricate(:user) }
+  let!(:device_story_comment) { Fabricate(:device_story_comment) }
+  describe 'POST #create' do
+    WebMock.allow_net_connect!
```
matschaffer, line 10: I'd take this out and use the calls in https://github.com/joshfrench/rakismet#testing to ensure rakismet doesn't actually try to check spam during testing.

matschaffer (changes requested):

app/models/device_story_comment.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class DeviceStoryComment < ApplicationRecord
+  include Rakismet::Model
+  validates :content, presence: true, length: { maximum: 1000 }
+  validates :user_id, presence: true
+  validates :device_story_id, presence: true
+  belongs_to :device_story
+  belongs_to :user
```
eitoball, line 9: It might be better to use `require` options instead of validating ids explicitly.

```suggestion
  belongs_to :device_story, require: true
  belongs_to :user, require: true
```

spec/requests/device_story_comments_controller_spec.rb
eitoball: I would use WebMock to check if it sends request to Akismet.

For example:

```ruby
let(:addressable) { Addressable::Template.new('https://{key}.rest.akismet.com/1.1/comment-check') }
...
before { stub_request(:post, addressable) }
...
  expect(WebMock).to have_requested(:post, addressable)
```

spec/requests/device_story_comments_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur
+        sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam
+        erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita
+        kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.Lorem ipsum dolor sit amet,
+        consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna
+        aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
+        clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor
+        sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore
+        magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum.
+        Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.',
```
eitoball, line 35: It might be easier to generate content with more than 1,000 characters.

```suggestion
        content: 'a' * 1_001
```

eitoball:
In spec, there are some lines that specify IDs or such like [here](https://github.com/Safecast/safecastapi/pull/743/files#diff-8c7dce4228d61df694f119f24edfebd50aab0883fe03947126c80abdc0063dbbR5-R6). It might be better not to specify IDs.

The diff below is a sample that tries not to specify IDs with modifications that I commented. I thought that showing code might be easier to understand.

```diff
diff --git a/spec/fabricators/device_story_comment_fabricator.rb b/spec/fabricators/device_story_comment_fabricator.rb
index db89db59..25588d86 100644
--- a/spec/fabricators/device_story_comment_fabricator.rb
+++ b/spec/fabricators/device_story_comment_fabricator.rb
@@ -2,6 +2,4 @@
 
 Fabricator(:device_story_comment) do
   content 'This is a test!'
-  user_id 1
-  device_story_id 1
 end
diff --git a/spec/requests/device_story_comments_controller_spec.rb b/spec/requests/device_story_comments_controller_spec.rb
index 2db5dd01..f6f27626 100644
--- a/spec/requests/device_story_comments_controller_spec.rb
+++ b/spec/requests/device_story_comments_controller_spec.rb
@@ -3,47 +3,35 @@
 require 'spec_helper'
 
 RSpec.describe DeviceStoryCommentsController, type: :request do
-  let!(:device_story) { Fabricate(:device_story, last_location_name: 'Israel', device_urn: 'safecast:114699387') }
-  let!(:user) { Fabricate(:user) }
-  let!(:device_story_comment) { Fabricate(:device_story_comment) }
+  let(:user) { Fabricate(:user) }
+  let(:device_story) { Fabricate(:device_story, last_location_name: 'Israel', device_urn: 'safecast:114699387') }
+
   describe 'POST #create' do
-    WebMock.allow_net_connect!
+    let(:addressable) { Addressable::Template.new('https://{key}.rest.akismet.com/1.1/comment-check') }
+
+    before { stub_request(:post, addressable).to_return(body: 'false') }
+
     it 'creates valid comment' do
-      comment_params = {
-        content: 'This is a test!',
-        device_story_id: DeviceStory.first.id,
-        user_id: User.first.id
-      }
-      expect { post '/device_stories/1/device_story_comments', params: { device_story_comment: comment_params } }.to change { DeviceStoryComment.count }.by(1)
+      comment_params = { content: 'This is a test!', user_id: user.id }
+      expect { post device_story_device_story_comments_path(device_story_id: device_story.id), params: { device_story_comment: comment_params } }.to change { DeviceStoryComment.count }.by(1)
       expect(response.status).to eq 302
+      expect(WebMock).to have_requested(:post, addressable)
     end
 
     it 'creates invalid comment' do
-      comment_params = {
-        content: 'Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam
-        nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua.
-        At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea
-        takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor sit amet, consetetur
-        sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam
-        erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita
-        kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.Lorem ipsum dolor sit amet,
-        consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna
-        aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet
-        clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet. Lorem ipsum dolor
-        sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore
-        magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum.
-        Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.',
-        device_story_id: DeviceStory.first.id,
-        user_id: User.first.id
-      }
-      expect { post '/device_stories/1/device_story_comments', params: { device_story_comment: comment_params } }.to change { DeviceStoryComment.count }.by(0)
+      comment_params = { content: 'a' * 1_001, user_id: user.id }
+      expect { post device_story_device_story_comments_path(device_story_id: device_story.id), params: { device_story_comment: comment_params } }.not_to change { DeviceStoryComment.count }
       expect(response.status).to eq 302
+      expect(WebMock).not_to have_requested(:post, addressable)
     end
   end
 
   describe 'DELETE #destroy' do
+    let(:device_story_comment) { Fabricate(:device_story_comment, device_story: device_story, user: user) }
+
     it 'deletes comment' do
-      expect { delete '/device_stories/1/device_story_comments/1' }.to change { DeviceStoryComment.count }.by(-1)
+      delete device_story_device_story_comment_path(device_story_id: device_story.id, id: device_story_comment.id)
+      expect { device_story_comment.reload }.to raise_error(ActiveRecord::RecordNotFound)
       expect(response.status).to eq 302
     end
   end
```

Nakushi64:
@eitoball , thank you for helping me out. Any idea why I get the `NoMethodError` in rspec when I use `sign_in user`? I looked at the other spec files and they seem to utilise it without any issues.

spec/requests/device_story_comments_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe DeviceStoryCommentsController, type: :request do
```
eitoball, line 5: > Any idea why I get the NoMethodError in rspec when I use sign_in user?

```suggestion
RSpec.describe DeviceStoryCommentsController, type: :controller do
```

@Nakushi64 If changed like above, it should solve error.

`Devise::Test::ControllerHelpers` that defines `sigin_user` method will be included. It is defined in `spec/spec_helper.rb` (`config.include Devise::Test::ControllerHelpers, type: :controller`)

Nakushi64:
@matschaffer @eitoball Alright, I just finished applying all the suggestions and all tests are successful when I use my Akismet key. If there is still anything to fix let me know. Otherwise I can think we can test it on the dev server.

matschaffer:
We should be able to get the tests green with something like @eitoball suggested in https://github.com/Safecast/safecastapi/pull/743#discussion_r547620690

spec/requests/device_story_comments_controller_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  let(:device_story) { Fabricate(:device_story, last_location_name: 'Israel', device_urn: 'safecast:114699387') }
+
+  before do
+    sign_in user
+  end
+
+  describe 'POST #create' do
+    let(:addressable) { Addressable::Template.new('https://{key}.rest.akismet.com/1.1/comment-check') }
+
+    before { stub_request(:post, addressable).to_return(body: 'false') }
```
Nakushi64, line 16: I have it already here.

Nakushi64:
@matschaffer I am out of ideas. When I use my key it works but using these environment variables
`Safecast::Application.config.rakismet.key = ENV['AKISMET_API_KEY']` 
` Safecast::Application.config.rakismet.url = "https://#{ENV['DEFAULT_HOSTNAME']}/"` 
it gives the `Rakismet.key is not defined` error...

matschaffer:
I suspect it’s the library throwing from a guard expression. There’s likely some option to suppress it for testing.

robouden:
Wish I could help out Emre, but unfortunately I am not so familiar with the issue.

eitoball:
@Nakushi64 You could mock `Rakismet.key` call like `allow(Rakismet).to receive(:key).and_return('12345678')`. This should be placed in `before` clause, I think.

Nakushi64:
@eitoball It worked! Thank you so much!

PR #750, opened 2020-12-23
Fix not to cache script content for chromium-base browsers

### LINKED ISSUE #746: Back button yields javascript output after search on device stories
Here's a recording https://recordit.co/ZOAho9JOej

And a screenshot

Screen Shot 2020-12-14 at 6 38 35

1. go to https://api.safecast.org/en-US/device_stories
2. type in a search (hit enter or not, either way)
3. click a device row
4. click back

eitoball:
Fixes #746

This occurs in Google Chrome and MicroSoft Edge, I think. It is because those browsers caches script contents, and they associate cached contents with histories.

### DIFF
--- a/app/views/device_stories/_search_form.html.erb
+++ b/app/views/device_stories/_search_form.html.erb
@@ -1,5 +1,5 @@
-<%= form_tag(device_stories_path, remote: true, :method => "get", enforce_utf8: false,
-             class: 'navbar-form navbar-left') do %>
+<%= form_tag(device_stories_path, remote: true, method: :get, enforce_utf8: false,
+             class: 'navbar-form navbar-left', data: { type: 'script' }) do %>
   <%= hidden_field_tag :fullscr, params[:fullscr] %>
   <%= hidden_field_tag :per_page, params[:per_page] %>
   <div class="input-group">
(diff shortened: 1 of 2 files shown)


app/views/device_stories/_search_form.html.erb
```diff
@@ -1,5 +1,5 @@
-<%= form_tag(device_stories_path, remote: true, :method => "get", enforce_utf8: false,
-             class: 'navbar-form navbar-left') do %>
+<%= form_tag(device_stories_path, remote: true, method: :get, enforce_utf8: false,
+             class: 'navbar-form navbar-left', data: { type: 'script' }) do %>
```
matschaffer, line 2: Is this data tag special some how? Curious how this works.

eitoball: Yes.  `data: { type: 'script' }` will be rendered as `data-type="script"` in HTML. jquery-ujs [reads](https://github.com/rails/jquery-ujs/blob/master/src/rails.js#L116) this value and [pass](https://github.com/rails/jquery-ujs/blob/master/src/rails.js#L147) it to `jQuery.ajax` method option as `dataType`.

For `dataType` in `jQuery.ajax`, please refer [this document](https://api.jquery.com/jquery.ajax/).

matschaffer:
Nice! I'm still stratching my head a little at _why_ it works, but I'm happy it does :)

Thanks @eitoball!

matschaffer (approved):

Issue #741, opened 2020-12-04
Cache measurement count


matschaffer:
We tried this in https://github.com/Safecast/safecastapi/pull/678 but it doesn't seem to really do anything.

Quite common to see a few queries all stacked up, even from what appears to be the same puma instance.

```
 29449 | 00:07:17.335181        | puma: cluster worker 0: 8483 [current] | 172.31.30.232 | SELECT COUNT(*) FROM "measurements"
 13603 | 00:07:17.335183        | puma: cluster worker 0: 8483 [current] |               | SELECT COUNT(*) FROM "measurements"
 13606 | 00:07:17.335184        | puma: cluster worker 0: 8483 [current] |               | SELECT COUNT(*) FROM "measurements"
```

Probably best to set up a job that counts it on a regular basis (e.g., nightly) and stores it to it's own DB record.

PR #761, opened 2021-01-09
Allow only one thread can count measurements

eitoball:
[Test-and-set](https://en.wikipedia.org/wiki/Test-and-set)

### DIFF
--- a/app/controllers/measurements_controller.rb
+++ b/app/controllers/measurements_controller.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-class MeasurementsController < ApplicationController
+class MeasurementsController < ApplicationController # rubocop:disable Metrics/ClassLength
   include SwaggerBlocks::Controllers::Measurements
 
   class << self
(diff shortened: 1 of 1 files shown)


matschaffer:
I still feel unsure of this approach (we have at least two instances, a memcached or redis would
incur more costs).

Have you considered a regularly generated (hourly cron) db entry?

eitoball:
> I still feel unsure of this approach (we have at least two instances, a memcached or redis would
> incur more costs).

@matschaffer I realize that `Rails.cache` is memory cache and available only for one process. But, it should solve case mentioned in #741

> Have you considered a regularly generated (hourly cron) db entry?

A little, but I thought that storing one number into database is too much. How do you think about it such as table schema?

matschaffer:
> How do you think about it such as table schema?

I was imagining something like `rails generate model CounterCache name:string:index value:integer` could do the trick.

I'm happy to give what's in this PR a try if you think it'll help. So long we can do it without additional infrastructure spend (e.g., adding a redis/memcached).

matschaffer:
Noting that @nokton mentioned using `SELECT reltuples::bigint AS estimate FROM pg_class where relname='mytable';` for a the measurements `count(*)`

> plz don't worry about the accuracy thing - it will report lower than what it is, but that is true for whatever number we post as it is just a snapshot in time. Not sure if you're running Analyze but that maybe needed

matschaffer:
Can probably close in favor of https://github.com/Safecast/safecastapi/pull/792, but will see how it goes in prod anyway.

PR #760, opened 2021-01-06
Mobile friendly Airnote device story page

Nakushi64:
- URLs of a pattern such as https://api.safecast.org/airnote/note:dev:865284040122792 are routed to a new responsive device story page which include air quality diagrams instead of the cpm diagrams
- If the device urn is invalid a help page is displayed instead (still work in progress)
Screenshot 2021-01-08 at 10 25 20
Screenshot 2021-01-08 at 10 25 38

PR #763, opened 2021-01-14
Responsive Layout

Nakushi64:
- Updated to the newest bootstrap-sass gem version: Bootstrap 2 -> Bootstrap 3
- Responsive layout for the whole API website
- 2 different navbars depending on the size of the window
- Includes changes made for #760 

Screenshot 2021-01-14 at 14 02 50
Screenshot 2021-01-14 at 14 03 17


app/helpers/device_stories_helper.rb
```diff
@@ -4,13 +4,16 @@ module DeviceStoriesHelper
...
+      air_quality: 15,
+      air_quality_map: 21
     }
     panels[name]
   end
 
-  def grafana_url(variant = nil)
-    url = ENV['DEVICE_STORIES_GRAFANA_BASE_URL'] || 'https://grafana.safecast.cc/d/DFSxrOLWk/safecast-device-details'
+  def grafana_url(variant = nil, is_airnote = false)
+    url = if is_airnote then ENV['DEVICE_STORIES_GRAFANA_BASE_URL'] || 'https://grafana.safecast.cc/d/7wsttvxGk/safecast-airnote'
```
matschaffer, line 15: `is_airnote` seems wrong here. Wonder if maybe we should move the `7wsttvxGk/safecast-airnote` vs `DFSxrOLWk/safecast-device-details` distinction into `panels`

app/views/bgeigie_imports/not_approved/index.html.haml
```diff
@@ -1,7 +1,7 @@
 = content_for(:title) { t('.bgeigie_imports') }
 
 %h2= t('.not_approved')
-
+%hr
```
matschaffer, line 4: How did you end up adding these? Just noticed it while browsing the site?

Nakushi64: I find they make the website visually more pleasing.

app/assets/stylesheets/application/safecast.scss
```diff
@@ -466,7 +488,7 @@ ul.measurement-listed {
 }
 
 .bootstrap-datetimepicker-widget {
-  z-index: 10000;
+  z-index: 10000 !important;
```
matschaffer, line 491: Can you provide a little more info on why we needed important here and on `#mobile-navbar`?

app/views/users/show.html.erb
```diff
@@ -1,20 +1,22 @@
 <%= content_for(:title) { "#{confirmed_name(@user)} \\\\ #{t('users.index.users')}" } %>
 
 <h2><%= t(".user") -%></h2>
+<hr/>
+<div class="well">
+  <dl class="dl-horizontal">
+    <dt style="width: 200px; margin-right: 10px"><%= t('id') %></dt>
```
matschaffer, line 7: Seems like we should probably make `width: 200px; margin-right: 10px` a class, or maybe you're setting the stage for utility CSS here?

app/assets/stylesheets/application/safecast.scss
Nakushi64: Leftovers from trying to solve an overlapping issue with the datetime pickers. Will take them out.

app/views/users/show.html.erb
matschaffer: re-up on this comment regarding the inline style

matschaffer (commented):
Looks pretty good overall. Could merge as-is but still had a few comments.

app/controllers/device_stories_controller.rb
```diff
@@ -25,6 +25,11 @@ def show
     respond_with @device_story
   end
 
+  def show_airnote
+    @device_story = DeviceStory.where(device_urn: params[:device_urn]).first
+    respond_with @device_story
```
eitoball, line 30: Using `#find_by!` might be better in case when `param[:device_urn]` is invalid. And, it raises exception.

```suggestion
    respond_with DeviceStory.find_by!(device_urn: params[:device_urn])
    respond_with @device_story
  rescue ActiveRecord::RecordNotFound
    head :not_found
  end
```

PR #771, opened 2021-02-12
Fix version of cryptography to 3.3.2

eitoball:
Since 3.4, it uses rust compiler.

To fix [this problem](https://safecast.slack.com/archives/C026F2Q1U/p1613092646150100).


matschaffer (approved):
Nice move :)

PR #777, opened 2021-02-19
Create container image for AWS CLI and eb cli

eitoball:
This PR create separate container image for AWS CLI and eb CLI.

To pass credential to container, you should use aws-vault. For example,

```
$ aws-valut exec safecast -- env | grep --color=no '^AWS_' > aws-cli.env
$ docker-compose --build aws-cli
...
```


matschaffer (approved):
I'll probably just stick to pip installing, but this PR won't impact me. If it makes your work easier, I'm happy to have it.

PR #802, opened 2021-04-17
Resolve deprecation warnings

eitoball:


config/environments/production.rb
```diff
@@ -46,7 +48,21 @@
   # config.action_cable.allowed_request_origins = [ 'http://example.com', /http:\/\/example.*/ ]
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  # config.force_ssl = true
+  config.force_ssl = true
+
+  config.ssl_options = {
+    redirect: {
+      exclude: proc do |request|
+        Rails.logger.error { request }
```
matschaffer, line 56: Why the logger here? Also the proc is a little curious. Seems like a controller method is easier. What prompted the switch?

matschaffer: The error logger is the main thing holding me up here. I'm fairly certain the pointcast gateway box comes in without https so we'll end up logging that as an error every time one of those (continual) sensors sends data.
```suggestion
```

matschaffer (approved):
Nice. I’m happy with this once it passes. But unfortunately we can’t deploy master to prod until we fix the device story pages.

config/environments/production.rb
eitoball: controller level `force_ssl` will be deprecated in rails 6.1, so we need to migrate to `ssl_options`. This `logger` was for debugging and should not be here, hence removed.

eitoball:
> Nice. I’m happy with this once it passes. But unfortunately we can’t deploy master to prod until we fix the device story pages.

@matschaffer Do you mean [this issue](https://github.com/Safecast/safecastapi/issues/806) ?

matschaffer:
Yep. That’s the one.

matschaffer (approved):
Looks solid. Thanks for removing the ssl redirect logging line.

PR #807, opened 2021-04-26
Fix undefined method in device stories page

eitoball:
Deployed to dev.safecast.org.


matschaffer (approved):
Thanks for the fix!

One thing this has me thinking is maybe we should get some testing around this that involves the elasticsearch container.

Since we're just passing in json, we could hit issues on our next elasticsearch upgrade and miss it in the test suite.

Anyway, I think this PR is great as is and unblocks master. So let's merge it.

eitoball:
> Anyway, I think this PR is great as is and unblocks master. So let's merge it.

@matschaffer How do we merge branch lately? "Rebase and merge" ?

matschaffer:
I’m fine with anything really. I usually squash and merge since it makes revert easy.

PR #810, opened 2021-05-08
Gitpod integration

matschaffer:
Hit https://gitpod.io/#https://github.com/matschaffer/safecastapi/tree/gitpod-integration to try it out

There are a couple of tweaks to watch out for:

1. I removed the aws-cli.env
  @eitoball or @robouden if you need that for local work, you can re-add it in a `docker-compose.override.yml` file. Leaving it out means we don't have to have an empty file for docker-compose to work (which creates dirty git state).

1. I switched the docker-compose ports to listen on any interface
  Back around 2017 I wanted to use 127.0.0.1 to avoid people accidentally having these listen on an insecure network. I think today such networks are less common and it's easier to leave OS firewalls enabled. Listening on any interface makes the docker-compose setup compatible with gitpod.io and probably generally easier for docker setups other than macos.

1. I removed the safecast_development setup from the postgis docker container
  I think this was causing some local dev issues too and we all just worked around it. I often find myself dropping the db, and recreating and I'm pretty sure this is why. If you let the container create the DB, postgis ends up in the wrong schema and causes the db:structure:load step to fail.

And about the rails plugin error you'll see if using VS Code on gitpod (default), I've raised that on their forums: https://community.gitpod.io/t/cannot-activate-the-rails-extension-error/3562

Things still basically work. I suspect we're just missing some rails-specific niceties.


eitoball (approved):

matschaffer:
Thanks @eitoball !

PR #819, opened 2021-06-01
Add current map snapshot export scripts

Frangible:
Not sure what best current way of executing this is.  The shell script will invoke the .sql scripts and generate an output file with psql and sqlite3.


matschaffer (approved):
Doesn't look like `dump_measurements_ios_periods.sh` is on a schedule, so someone can just run it from `eb ssh ... wrk` after deploy.

Issue #794, opened 2021-03-30
Note device landing page enhancements


matschaffer:
Opening this for tracking improvements on https://api.safecast.org/airnote/note:dev:864475040529876 before switching the airnote.live links to point directly to safecast pages rather than grafana.

- change time ranges directly
- see battery history
- less emphasis on location (you already know because it's yours)
- Ensure Radiation/Air graphs are selected appropriately for air/rad notes

PR #824, opened 2021-06-20
add time range selector to device story page

ZacharyCChang0828:
Add a time range selector to device story page

Rel: https://github.com/Safecast/safecastapi/issues/794

### DIFF
--- a/app/assets/javascripts/device_stories.js
+++ b/app/assets/javascripts/device_stories.js
@@ -1,0 +1,11 @@
+window.addEventListener("load", () => {
+  const url = new URL(document.location);
+
+  document.querySelectorAll("[data-behavior='change-time-range']").forEach((el) => {
+    const timeRange = el.dataset.timeRange;
+    el.addEventListener("click", (e) => {
+      e.preventDefault();
+      history.pushState({ time_range: timeRange }, el.innerText, `${url.pathname}?time_range=${timeRange}`);
+    });
+  })
+});
--- a/app/controllers/device_stories_controller.rb
+++ b/app/controllers/device_stories_controller.rb
@@ -22,16 +22,7 @@ class DeviceStoriesController < ApplicationController
   end
 
   def show
-    if params.key?(:time_range)
-      @device_story.time_range = params[:time_range]
-      respond_to do |format|
-        format.html
-        format.json { render :show, location: @device_story }
-        format.js { render js: 'window.top.location.reload(true);' }
-      end
-    else
-      respond_with @device_story
-    end
+    respond_with @device_story
   end
 
   def show_airnote
--- a/app/controllers/device_stories_controller.rb
+++ b/app/controllers/device_stories_controller.rb
@@ -26,6 +26,11 @@ class DeviceStoriesController < ApplicationController
                     else
                       DeviceStory.find_by!(device_urn: params[:device_urn])
                     end
+    if params.key?(:time_range)
+      @device_story.time_range = params[:time_range]
+    else
+      @device_story.time_range = 1
+    end
     respond_with @device_story
   rescue ActiveRecord::RecordNotFound
     head :not_found
--- a/app/controllers/device_stories_controller.rb
+++ b/app/controllers/device_stories_controller.rb
@@ -4,6 +4,7 @@ class DeviceStoriesController < ApplicationController
   has_scope :order
 
   before_action :authenticate_user!, only: %i(new create edit update destroy)
+  before_action :fetch_device_story, only: %i(show)
 
   layout :current_layout
 
@@ -21,26 +22,16 @@ class DeviceStoriesController < ApplicationController
   end
 
   def show
-    @device_story = if params.key?(:id)
-                      DeviceStory.find(params[:id])
-                    else
-                      DeviceStory.find_by!(device_urn: params[:device_urn])
-                    end
     if params.key?(:time_range)
-      puts'has time range'
       @device_story.time_range = params[:time_range]
-      puts @device_story.time_range
       respond_to do |format|
-        format.html 
-        format.json { render :show, location: @device_story}
+        format.html
+        format.json { render :show, location: @device_story }
         format.js { render js: 'window.top.location.reload(true);' }
       end
     else
-      puts @device_story.time_range
       respond_with @device_story
     end
-  rescue ActiveRecord::RecordNotFound
-    head :not_found
   end
 
   def show_airnote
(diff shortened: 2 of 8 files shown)


app/views/device_stories/_time_range_select.erb
```diff
@@ -0,0 +1,3 @@
+<%= form_tag controller_name ,method: :get, id: 'time_range_form' do %>
```
matschaffer, line 1: @eitoball or @ma2gedev might be good folks to weigh-in here, but pretty sure you can write this as `form_with` on `@device_story` to get back to the same path.

app/models/ingest_measurement.rb
```diff
@@ -12,14 +12,14 @@ def data_for(query)
       search(query: query).results.map(&:_source)
     end
 
-    def query_sensor_data(device_urn) # rubocop:disable Metrics/MethodLength
+    def query_sensor_data(device_urn,time_range) # rubocop:disable Metrics/MethodLength
```
eitoball, line 15: Maybe use default value like

```suggestion
    def query_sensor_data(device_urn, time_range = 'now-8w') # rubocop:disable Metrics/MethodLength
```

so that you won't have to change line 14 of `app/views/device_stories/show.html.erb`.

app/views/device_stories/_time_range_select.erb
eitoball:
```suggestion
<%= form_with model: @device_story, method: :get, id: 'time_range_form' do %>
```

app/models/ingest_measurement.rb
ZacharyCChang0828: Thanks!

app/views/device_stories/_time_range_select.erb
ZacharyCChang0828: Thank you! that worked like a charm.

ZacharyCChang0828:
Thanks for the advice on the form! I just made another PR. I am stuck on the implementation on how to pass the parameter back to the page itself. 

From my understanding now. time_range is now getting passed to the controller devicestory#show method which calls a respond_with @deivce_story.

I tried to add a :time_range field to the @device_story model but it is trickier than I thought, I wonder what's the right way to do it.

I also wonder if there is a way to pass parameter through repond_with or respond_to which can be another possible implementation.

app/models/device_story.rb
```diff
@@ -3,6 +3,7 @@
 class DeviceStory < ApplicationRecord
   validates :device_urn, uniqueness: true, presence: true
   has_many  :device_story_comments, dependent: :destroy
+  has_one   :time_range
```
ZacharyCChang0828, line 6: attempt to add a new field

app/controllers/device_stories_controller.rb
```diff
@@ -26,6 +26,11 @@ def show
                     else
                       DeviceStory.find_by!(device_urn: params[:device_urn])
                     end
+    if params.key?(:time_range)
+      @device_story.time_range = params[:time_range]
+    else
+      @device_story.time_range = 1
+    end
     respond_with @device_story
```
ZacharyCChang0828, line 34: set time range value based on :time_range passed through form

app/controllers/device_stories_controller.rb
```diff
@@ -26,6 +26,11 @@ def show
                     else
                       DeviceStory.find_by!(device_urn: params[:device_urn])
                     end
+    if params.key?(:time_range)
+      @device_story.time_range = params[:time_range]
+    else
+      @device_story.time_range = 1
+    end
```
matschaffer, line 33: `@time_range = params[:time_range]`

app/views/device_stories/show.html.erb
```diff
@@ -9,9 +9,15 @@
     <%- end %>
   </h2>
 </div>
+<div id="time_range_select">
+  <%= render 'time_range_select'%>
+  <%- if @device_story.time_range == 2 %>
```
matschaffer, line 14:
```suggestion
  <%- if @time_range == 2 %>
```

matschaffer (commented):
Comments from meeting

ZacharyCChang0828:
The code above made it possible to make selection to time_range through directly adding ?time_range=2 after the url at each device story page. I am currently trying to find a waay to pass parameter through respond_with. I did stumble across the attr_accessor which is capable of adding a temporary parameter to @deivce_story obeject for passing parameter. However, I wonder if any change has to be done to the controller and routes.rb to make the form work so the page renders the page again with the right parameter as the form gets submitted. I tried to use respond_to but the parameter got lost as it tries to reload the page.

eitoball:
It should be working now. I added JavaScript to fetch page with time range parameter.

I changed to use buttons instead of select options. I can change back to select options anytime.

Screen Shot 2021-11-05 at 0 21 40

app/assets/javascripts/device_stories.js
```diff
@@ -0,0 +1,10 @@
+window.addEventListener("load", function (e) {
+  ["1", "2", "3"].forEach(function (selection) {
```
matschaffer, line 2: Seems like we could make this a little clearer than 1,2,3

matschaffer (commented):
Guessing this needs conflict resolution before we propose merging

app/assets/javascripts/device_stories.js
eitoball: How about this? Changed to use `id` value directory.

matschaffer: It's a little better.

I'm wondering why it's not more like https://www.bigbinary.com/blog/data-behavior#data-behavior-to-the-rescue though - seems like basically we're just binding a behavior to navigate to the currently selected param.

That might allow for adding/removing timeframes without potentially breaking the JS.

eitoball: Changed. How about now?

app/views/device_stories/show.html.erb
```diff
@@ -9,11 +8,20 @@
     <%- end %>
   </h2>
 </div>
+<div class="row">
+  <div class="col-sm-12 text-right">
+    <div class="btn-group" role="group" aria-label="Time Range">
+      <button class="btn btn-sm <%= params[:time_range].blank? || params[:time_range] == "1" ? "btn-primary" : "btn-default" -%>" data-behavior="change-time-range" data-time-range="1">2 months</button>
+      <button class="btn btn-sm <%= params[:time_range] == "2" ? "btn-primary" : "btn-default" -%>" data-behavior="change-time-range" data-time-range="2">6 months</button>
+      <button class="btn btn-sm <%= params[:time_range] == "3" ? "btn-primary" : "btn-default" -%>" data-behavior="change-time-range" data-time-range="3">1 year</button>
```
matschaffer, line 16: Still not a big fan of the 1/2/3 stuff going on here, but at least now it's all in the ruby view layer (before it was also in the JS layer).

matschaffer (approved):
overall LGTM. We can polish it later if it's ever a problem for future maintenance.

PR #832, opened 2021-10-24
Change /measurements not to issue count query

### LINKED ISSUE #811: Remove kaminari table count from /measurements
Based on some recent DB analysis, we have some DB load still coming from /measurements

here's an example trace in APM: https://5bc3d4f7330a4459881197a85659caf5.us-west-2.aws.found.io:9243/app/apm#/services/safecastapi-prd/transactions/view?rangeFrom=now-24h&rangeTo=now&refreshPaused=true&refreshInterval=0&kuery=&traceId=6a1ae7c5cdfc17cac4185d215792595e&transactionId=42cefa2d20c2ac8b&transactionName=MeasurementsController%23index&transactionType=request&flyoutDetailTab=&waterfallItemId=

it shows 4.1min spent on `SELECT COUNT(*) FROM "measurements"` starting from `layouts/_table_sort_header.html.erb` then going into a bunch of kaminari erb.

Seems like we need to swich that to an estimate, similar to what we did with MeasurementsController#count https://github.com/Safecast/safecastapi/blob/main/app/controllers/measurements_controller.rb#L106

Or maybe there's some smarter way we can use kaminari to just not try to give you a full page count.

eitoball:
Fixes #811 

# Screenshots

## First page

Screen Shot 2021-10-24 at 17 59 58

## Middle page

Screen Shot 2021-10-24 at 18 00 06

## Last page

Screen Shot 2021-10-24 at 18 00 35

### DIFF
--- a/app/assets/stylesheets/application/safecast.scss
+++ b/app/assets/stylesheets/application/safecast.scss
@@ -643,4 +643,34 @@ i.muted {
   padding: 1px;
   text-align: center;
   background: #FFFF00;
-} 
+}
+
+.kaminari-pager {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-top: 1rem;
+  padding: 0.4rem 1.6rem;
+
+  &__item {
+    letter-spacing:0.1rem;
+
+    &__prev {
+      &::before {
+        content: "<<";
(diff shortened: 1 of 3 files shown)


matschaffer (approved):
Still LGTM

PR #833, opened 2021-10-24
Prevent highchart.js from getting source map file

eitoball:
This PR should prevent highchart.js from getting its source map file. This causes error like:

```
app_1            | ActionController::RoutingError (No route matches [GET] "/assets/highcharts.js.map"):
```

See [this thread](https://www.highcharts.com/forum/viewtopic.php?t=41642)


matschaffer (approved):
Neat. Seems worth a shot!

PR #834, opened 2021-10-30
KMZ download

### LINKED ISSUE #758: KMZ file download
See #752 

Add option to download KMZ file; a file with KML and icon files.

eitoball:
Fixes #758


matschaffer (approved):
Gonna just stamp this one. Easy enough to tweak if we need to.

PR #835, opened 2021-11-03
Create configuration file for CodeQL

eitoball:
[CodeQL](https://codeql.github.com) seems to be available as *beta*. I would like to try this to see if CodeQL helps to find vulnerabilities in our code.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,5 @@
+name: "CodeQL config"
+
+paths-ignore:
+  - vendor/assets
+  - app/assets/javascripts/bgeigie_viewer.js
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,71 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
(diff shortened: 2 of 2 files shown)


matschaffer (approved):

matschaffer:
On thing that comes to mind is that we might not want to block merges on this until we have an idea what it'll show us. I'm good to try it though.

matschaffer:
oh nice, it already analyzed this PR :)

matschaffer (changes requested):
Gonna take back my approve given what's on https://github.com/Safecast/safecastapi/pull/835/checks?check_run_id=4339856558

Let's figure out how to at least get rid of the flags in the vendored JS assets (ignore is fine) before we put this on main, otherwise builds will probably get very noisy.

.github/codeql/codeql-config.yml
```diff
@@ -0,0 +1,4 @@
+name: "CodeQL config"
+
+paths-ignore:
+  - vendor/assets
```
matschaffer, line 4: We can add the bgeigie viewer to this list as well. It's sourced from https://github.com/Safecast/Tilemap/blob/master/bgeigie_viewer.js but I didn't find a better way to include it than just checking it in.

matschaffer (approved):
Looking good! Any idea what those XSS warnings are referring to? It's not clear what we're doing wrong there.

PR #836, opened 2021-11-03
Refactor Sign-up and Register forms

eitoball:
## Sign-in form

Screen Shot 2021-11-03 at 15 10 49

## On mobile view (iPhone SE)

Screen Shot 2021-11-03 at 15 11 44

## Register form

Screen Shot 2021-11-03 at 15 11 01


matschaffer (approved):
Great work!

PR #837, opened 2021-11-03
Hide non-moderator E-mail address when rejecting own bGeigie import

### LINKED ISSUE #797: Hide non-moderator email addresses from rejection information
We noticed recently that if people self-reject drives the user's email gets copied into the `rejected_by` column which is then available via the public API:

```
~                                                                                                                                                                                                                  15:25:46
❯ curl -s 'https://api.safecast.org/en-US/bgeigie_imports/47672?format=json&id=47672' | jq -r .rejected_by
h...@gmail.com
```

We should probably look at hiding this information to avoid leaking people's email addresses.

eitoball:
Fixes #797

If you reject your own drive, it would fill with`<self>` instead of E-mail address.


.elasticbeanstalk/package.sh
```diff
@@ -26,7 +26,7 @@ CLEAN_DESCRIPTION="${DESCRIPTION/https*:\/\//}"
 
 GIT_COMMIT="$(git rev-parse --verify HEAD)"
 
-VERSION="${EB_APP_NAME}-${CLEAN_BRANCH_NAME}-${BUILD_NUMBER}-${GIT_COMMIT}"
+VERSION="${EB_APP_NAME}-${CLEAN_BRANCH_NAME:0:32}-${BUILD_NUMBER}-${GIT_COMMIT}"
```
matschaffer, line 29: ha! Guess we exceeded the limit here

matschaffer (approved):

PR #839, opened 2021-11-05
Download Postgres RPMs by  curl with insecure option 

eitoball:
This Is a temporary solution.

We should migrate to Amazon Linux 2. See [this page](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features.migration-al.html?icmpid=docs_elasticbeanstalk_console).

---

Deployed to safecastapi-dev-032 and safecastapi-dev-wrk-032, they seem to be fine.

### DIFF
--- a/.ebextensions/postgresql.config
+++ b/.ebextensions/postgresql.config
@@ -2,10 +2,16 @@ commands:
   upgrade_postgres:
     command: |
       /usr/bin/yum -y remove 'postgres*'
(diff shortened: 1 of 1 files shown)


matschaffer (approved):
No problem with this as a temp measure while we look at platform upgrade.

eitoball:
Closing this PR since #840

PR #844, opened 2021-12-01
Update elastic-apm from 3.6.0 to 4.4.0

eitoball:
To fix errors like following.

```
E, [2021-11-30T15:35:08.958025 #29637] ERROR -- : [ElasticAPM] Failed converting event to JSON: <ElasticAPM::Metricset timestamp:1638286508948138 transaction:nil span:nil tags:nil samples:{:"system.cpu.total.norm.pct"=>NaN, :"system.memory.actual.free"=>1013149696, :"system.memory.total"=>2087239680, :"system.process.cpu.total.norm.pct"=>NaN, :"system.process.memory.size"=>773079040, :"system.process.memory.rss.bytes"=>127246336}>
```


matschaffer (approved):
Nice! I'd like to also upgrade the backend elasticsearch some time soon too. 7.15,7.16 has some nice features for correlating slow APM traces that might help us.

Gemfile
```diff
@@ -54,6 +53,10 @@ gem 'tzinfo-data', platforms: %i(mingw mswin x64_mingw)
 gem 'uglifier'
 gem 'wdm', platforms: %i(mingw mswin x64_mingw)
 
+group :production do
```
matschaffer, line 56: Ah, one thing is here is that I was keeping it in dev intentionally so we could use it at dev-time as well. Guess either way is fine though since most people wouldn't go get the APM token for their dev env.

PR #846, opened 2021-12-05
Prevent from running multiple same cron jobs

eitoball:
This PR is trying to prevent cron jobs from using database storages.

### DIFF
--- a/app/controllers/crons_controller.rb
+++ b/app/controllers/crons_controller.rb
@@ -4,18 +4,6 @@ class CronsController < ApplicationController
   SCRIPT_PATH = Rails.root.join('cron')
   CRON_DEFINITIONS = Rails.root.join('cron.yaml')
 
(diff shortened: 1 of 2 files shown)


app/controllers/crons_controller.rb
```diff
@@ -27,4 +15,31 @@ def create
...
+    YAML.load_file(CRON_DEFINITIONS)['cron'].map { |d| d['name'] }
+  end
+
+  def taskname
+    @taskname ||= request.headers['X-Aws-Sqsd-Taskname']
+  end
+
+  def running_same_job?
+    res = false
+    IO.popen('ps ax') do |io|
```
matschaffer, line 35: This won't work beyond one worker. Though we've only ever had one so probably fine.

matschaffer (approved):

Issue #782, opened 2021-03-05
Load in cs-CZ translation


matschaffer:
Apparently [Jan](https://twitter.com/jhelebrant/status/1367814863372877824?s=20) added all the cs-CZ translations to https://app.localeapp.com/projects/4489/translations?in_locale=27959

Seems like there should be some way to get this back in the app for more localizations beyond english and japanese.

matschaffer:
Looks like it got removed back in https://github.com/Safecast/safecastapi/commit/799b01c5b22c9c74e3fa162e6b659d4e739953a6 - nothing saying why.

Not sure if we want to continue using it, but it'd be good to at least commit the translations from Jan to git so we can leverage his work.

eitoball:
It is on api.safecast.org now.

PR #847, opened 2021-12-07
Add czech and portuguese translations to locales

GuiHeurich:
*why*

There are translations in https://app.localeapp.com/projects/4489 which could be used. 

*what*

Add yaml files with translations in Czech and Portuguese to config/locales.

Rel: https://github.com/Safecast/safecastapi/issues/782

### DIFF
--- a/config/locales/cs-CZ.yml
+++ b/config/locales/cs-CZ.yml
@@ -1,0 +1,490 @@
+cs-CZ:
+  My_maps: Moje mapy
+  activerecord:
+    attributes:
+      bgeigie_import:
+        created_at: Nahráno dne
+        measurements_count: "# měření"
+        q: q
+        q_tooltip: Hledat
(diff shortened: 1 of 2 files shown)


matschaffer (approved):
Looks great! how did you get the export?

GuiHeurich:
There is a "downloads" tab somewhere on localeapp.

matschaffer:
awesome! great find. I'll merge this for now. Do we need to update the UI to allow selection of the new locale as well?

GuiHeurich:
I haven't tested that, actually, but will have a look.

matschaffer:
Thanks! Hopefully it's easy and we can do it as part of https://github.com/Safecast/safecastapi/issues/782 - if not we can close that one and open another issue to add the locale selector for the UI.

PR #848, opened 2021-12-10
Update ruby to 2.7.6

### LINKED ISSUE #646: Upgrade api to Ruby 2.7

### LINKED ISSUE #648: Customize api nginx to rate-limit bots

eitoball:
Fixes #646

[Elastic Beanstalk supported platforms - Ruby](https://docs.aws.amazon.com/elasticbeanstalk/latest/platforms/platforms-supported.html#platforms-supported.ruby)


matschaffer (approved):
Nice! Of course, let's check on dev first.

matschaffer (approved):
Still looks good. Happy new year @eitoball !

PR #852, opened 2021-12-18
Fix time range links in device story page

eitoball:
This PR fixes time range links in device story page that are added in #824.


matschaffer (approved):
Didn't run it, but seems reasonable :)

PR #854, opened 2021-12-20
Reset to default locale

### LINKED ISSUE #853: Address flakey DeviceStoryComment spec
See https://app.circleci.com/pipelines/github/Safecast/safecastapi/1063/workflows/723033c2-d0bf-4e82-9102-ead2378dfa3b/jobs/1563/tests#failed-test-0 for example failure.

This test has been failing occasionally during runs from main and PRs.

eitoball:
Fixes #853

When locale is `:ja`, space between attribute name and message is removed.

TODO: i18n-ize error messages in `DeviceStoryComment`

### DIFF
--- a/spec/models/device_story_comment_spec.rb
+++ b/spec/models/device_story_comment_spec.rb
@@ -9,8 +9,22 @@ RSpec.describe DeviceStoryComment, type: :model do
   describe '#save' do
     let(:comment) { described_class.new(image: upload_file, content: '_content_', user: user, device_story: device_story) }
 
+    def file_upload(src, content_type, binary: false)
+      path = Rails.root.join(src)
+      original_filename = path.basename.to_s
+
+      tempfile = Tempfile.new(original_filename)
+      binary ? tempfile.binwrite(path.binread) : tempfile.write(path.read)
+      tempfile.rewind
+
+      uploaded_file = Rack::Test::UploadedFile.new(tempfile, content_type, binary, original_filename: original_filename)
+      ObjectSpace.define_finalizer(uploaded_file, uploaded_file.class.finalize(tempfile))
+
+      uploaded_file
+    end
+
     context 'with JPEG image' do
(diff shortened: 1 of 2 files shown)


spec/models/device_story_comment_spec.rb
```diff
@@ -9,19 +9,35 @@
...
     end
 
     context 'with JSON file' do
-      let(:upload_file) { fixture_file_upload('files/device_stories.json') }
+      let(:upload_file) { file_upload('spec/fixtures/files/device_stories.json', 'application/json') }
 
       it 'should have attached image' do
         expect(comment.save).to be_falsey
+        # binding.pry
+        comment.errors.full_messages
```
matschaffer, line 40: Guessing these two lines don't need to be here right?

eitoball: Right. Thank you for catching.

spec/models/device_story_comment_spec.rb
```diff
@@ -9,16 +9,30 @@
   describe '#save' do
     let(:comment) { described_class.new(image: upload_file, content: '_content_', user: user, device_story: device_story) }
 
+    def file_upload(src, content_type, binary: false)
```
matschaffer, line 12: Why do we need our own upload function? Seems like https://github.com/rails/rails/blob/v6.0.3.7/actionpack/lib/action_dispatch/testing/test_process.rb#L17 supports mime type definition.

matschaffer: Ah, nm. I see

matschaffer, line 12:
```suggestion
    # Workaround for ActiveStorage::IntegrityError, see https://github.com/rails/rails/issues/41991
    def file_upload(src, content_type, binary: false)
```

matschaffer (changes requested):
Let's not the reason for the custom uploader just to avoid future confusion. Other than that LGTM

matschaffer (approved):

Issue #873, opened 2022-08-11
Measurements dump may be downloaded incomplete


KETHERCORTEX:
Every time the compressed database dump (`measurements.tar.gz`) is updated, previous file gets deleted at the beginning of the new file generation. 

If the file is downloaded during the generation process, user may end up with an incomplete not fully formed file. According to the server responding with increasing file size, `measurements.tar.gz` generation may take more than 15 minutes. And it takes less than 15 minutes to download full (~5.9GB right now) file.

KETHERCORTEX:
> According to the server responding with increasing file size

Log fragment from `wget` when connection was interrupted a few times illustrates that:

```
--2022-08-11 16:17:57--  (try: 8)  https://api.safecast.org/system/measurements.tar.gz
Length: 4688805888 (4,4G), 3659969845 (3,4G) remaining [application/octet-stream]

--2022-08-11 16:18:41--  (try: 9)  https://api.safecast.org/system/measurements.tar.gz
Length: 4821401600 (4,5G), 3583620405 (3,3G) remaining [application/octet-stream]

--2022-08-11 16:19:24--  (try:10)  https://api.safecast.org/system/measurements.tar.gz
Length: 5073076224 (4,7G), 3628184885 (3,4G) remaining [application/octet-stream]
```
_(only lines with date/time, connection attempt number and file size left, the rest is omitted)_

eitoball:
Thank you for your PR. It has been merged and deployed to production server.

PR #874, opened 2022-08-11
Non-conflicting dump archive naming #873

KETHERCORTEX:
Measurements `.csv` file gets fully generated at first with non-conflicting name (`measurements-out.csv`), and only then replaces an old one by renaming. Used the same approach for `tar.gz` file.


robouden:
Seems fine for me to merge if needed.

eitoball (approved):

PR #881, opened 2022-08-14
Use config/elastic_apm.yml instead

eitoball:
According to [this page](https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html),

> There are several ways to configure how Elastic APM behaves. We recommend using a config/elastic_apm.yml file:


eitoball:
Deployed to dev, and seems to be working fine.

PR #987, opened 2022-12-07
Fix error that occurs when specifying multiple condition in bgeigie_imports#index

eitoball:
This PR should fixes error. This fix will resolve issue that there is no data points when accessing https://map.safecast.org/?logids=qbgeigies4ukraine

Screen Shot 2022-12-07 at 21 09 25


robouden:
Thanks for the fix!!!

